PR #45, opened 2025-01-03
DuckDB: Improve SQL management to avoid injection (#44)

### LINKED ISSUE #44: DuckDB Embeddings Store has SQL injection vulnerabilty
When the text being encoded has a single quote the code breaks. This bug could also be used to to hack the store. See also
[https://xkcd.com/327/](https://xkcd.com/327/)

The assigment of the `text` variable below is not safe the insert will fail if `embedded.get(i).text()` contains a quote or newline or be used to faciliate a SQL injection attack.

```java
        try (var connection = duckDBConnection.duplicate();
                var statement = connection.createStatement()) {
            var values = new ArrayList<String>(ids.size());
            for (int i = 0; i < ids.size(); i++) {
                var text = "NULL";
                if (embedded != null && embedded.get(i) != null) {
                    text = "'" + embedded.get(i).text() + "'";
                }
                var metadata = embedded != null && embedded.get(i) != null
                        ? embedded.get(i).metadata().toMap()
                        : null;
                values.add(format(
                        "('%s',%s,%s,'%s')",
                        ids.get(i),
                        embeddingToParam(embeddings.get(i)),
                        text,
                        jsonMetadataSerializer.writeValueAsString(metadata)));
            }

            var sql = format(INSERT_QUERY_TEMPLATE, tableName, String.join(",", values));
            log.debug(sql);
            statement.execute(sql);
        } catch (SQLException | JsonProcessingException e) {
            throw new DuckDBSQLException("Unable to add embeddings in DuckDB", e);
        }
```

https://github.com/langchain4j/langchain4j-community/blob/062e4691e66029ab0b4d3633bf0340b665d3a1de/embedding-stores/langchain4j-community-duckdb/src/main/java/dev/langchain4j/community/store/embedding/duckdb/DuckDBEmbeddingStore.java#L269

fb64:
## Issue
Fixes #44

## Change
Use JDBC `PreparedStatement` instead of raw SQL to avoid injection in `DuckDBEmbeddingStore` 

## General checklist

### DIFF
--- a/embedding-stores/langchain4j-community-duckdb/src/main/java/dev/langchain4j/community/store/embedding/duckdb/DuckDBEmbeddingStore.java
+++ b/embedding-stores/langchain4j-community-duckdb/src/main/java/dev/langchain4j/community/store/embedding/duckdb/DuckDBEmbeddingStore.java
@@ -48,22 +48,21 @@ public class DuckDBEmbeddingStore implements EmbeddingStore<TextSegment> {
 
     private static final String SEARCH_QUERY_TEMPLATE =
             """
-            select *, (list_cosine_similarity(embedding,%s)+1.0)/2.0 as score
+            select id, embedding, text, metadata, (list_cosine_similarity(embedding,%s)+1.0)/2.0 as score
             from %s
             where score >= %s %s
             order by score DESC
-            limit(%d)
+            limit %d
             """;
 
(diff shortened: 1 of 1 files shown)


Martin7-1 (approved):
@fb64 Thank you!

PR #96, opened 2025-03-07
Support wanx v2 models and their new features

jiangsier-xyz:
## Change

Support wanx v2 models and their new features:
- wanx2.0-t2i-turbo: v2.0, cheaper than v2.1
- wanx2.1-t2i-turbo: v2.1, faster speed
- wanx2.1-t2i-plus: v2.1, better effect

## General checklist


Martin7-1 (approved):
@jiangsier-xyz Thank you!

Issue #111, opened 2025-03-24
[FEATURE] Neo4j Integration Improvements List


jexp:
As part of the Neo4j integration in langchain4j community we want to improve the base functionality.

## Neo4jGraph

- [ ] Switch to driver.executeQuery to automatically get retries
- [ ] Combine metadata retrieval into one query
- [ ] Add option for enhanced schema with sample property values
- [ ] Integration test for invalid query
- [ ] Fixing generated queries, e.g. relationship-direction (e.g. with [cypher-dsl](https://github.com/neo4j/jbang-catalog/blob/main/enforce_relationship_directions.java)) 

## Neo4jContentRetriever

- [ ] Rename to Neo4jText2CypherContentRetriever
- [ ] Re-generate with the question, query, error and re-run failed generated queries 
- [ ] Integration test for invalid query
- [ ] Add a more comprehensive integration test with a bit more complex schema e.g. northwind?

## Neo4j Vector Support

- [ ] Add Hybrid Search with re-ranker, fulltext index generation
- [ ] Add GraphRAG retrieval concepts
- [ ] Support parent-child and entity-retrievers

## Documentation & Examples

- [ ] Document the existing and new features in more detail
- [ ] Extend the example code repository with more practical examples

## Graph Construction

- [ ] LLMGraphTransformer equivalent
- [ ] Provide GraphSchema from user or from Neo4jGraph
- [ ] Prompt for entity extraction
- [ ] Graph representation as response
- [ ] Process Documents with LLM + Prompt + structured output
- [ ] Allow Neo4jGraph to store graph representation response directly and efficiently (batched)

## Graphs As Conversation / Agent Memory

- [ ] Integrate Neo4j as storage for conversation history / memory similar to LC
- [ ] Add agentic memory integration when possible

vga91:
Neo4jGraph PRs:
-  https://github.com/langchain4j/langchain4j-community/pull/119
- https://github.com/langchain4j/langchain4j-community/pull/129

Neo4jContentRetriever PR: 
- https://github.com/langchain4j/langchain4j-community/pull/120

Neo4j Vector Support PRs:
- https://github.com/langchain4j/langchain4j-community/pull/113
- https://github.com/langchain4j/langchain4j-community/pull/113
- TODO: parent-child retriever

Documentation & Examples's PRs
- https://github.com/langchain4j/langchain4j-examples/pull/155
- https://github.com/langchain4j/langchain4j/pull/2972
[...]

PR #147, opened 2025-04-23
Issue 111: GraphRAG retrieval concepts

vga91:
## Issue

Closes Add GraphRAG retrieval concepts of #111

## Change

- Improved Neo4jEmbeddingStore with entityCreationQuery and additionalParam configurations

Created inject since the EmbeddingStoreIngestor doesn't create entities and doens't support 2 splitters together.

Created promptAnswer to improve the retrieving, since we tried with a RetrievalAugmentor but it seems not to work correctly
```
RetrievalAugmentor retrievalAugmentor = DefaultRetrievalAugmentor.builder()
.contentRetriever(contentRetriever)
.contentInjector(...)
.queryRouter()
.queryTransformer(queryTransformer)
                .build();

return AiServices.builder(Assistant.class)
                .chatLanguageModel(chatLanguageModel)
                .retrievalAugmentor(retrievalAugmentor)
                .build();
```

## General checklist

### DIFF
--- a/content-retrievers/langchain4j-community-neo4j-retriever/src/test/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4jEmbeddingRetrieverTest.java
+++ b/content-retrievers/langchain4j-community-neo4j-retriever/src/test/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4jEmbeddingRetrieverTest.java
@@ -26,7 +26,7 @@ import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
 
 @ExtendWith(MockitoExtension.class)
-public class Neo4jEmbeddingRetrieverTest extends Neo4jEmbeddingRetrieverBaseTest {
+public class Neo4jEmbeddingRetrieverTest extends Neo4jEmbeddingStoreIngestorBaseTest {
 
     @Mock
     private ChatModel chatLanguageModel;
--- a/content-retrievers/langchain4j-community-neo4j-retriever/src/test/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4jEmbeddingStoreIngestorTest.java
+++ b/content-retrievers/langchain4j-community-neo4j-retriever/src/test/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4jEmbeddingStoreIngestorTest.java
@@ -6,7 +6,7 @@ import static org.junit.jupiter.api.Assertions.assertTrue;
 import static org.mockito.ArgumentMatchers.anyList;
 import static org.mockito.Mockito.when;
 
-import dev.langchain4j.community.embedding.ParentChildEmbeddingStoreIngestor;
+import dev.langchain4j.community.store.embedding.ParentChildEmbeddingStoreIngestor;
 import dev.langchain4j.community.store.embedding.neo4j.Neo4jEmbeddingStore;
 import dev.langchain4j.community.store.embedding.neo4j.Neo4jEmbeddingStoreIngestor;
 import dev.langchain4j.data.document.Document;
@@ -40,13 +40,13 @@ public class Neo4jEmbeddingStoreIngestorTest extends Neo4jEmbeddingStoreIngestor
         int maxSegmentSize = 250;
         DocumentSplitter splitter = new DocumentByRegexSplitter(expectedQuery, expectedQuery, maxSegmentSize, 0);
 
-        final ParentChildEmbeddingStoreIngestor build = ParentChildEmbeddingStoreIngestor.builder()
+        final ParentChildEmbeddingStoreIngestor ingestor = ParentChildEmbeddingStoreIngestor.builder()
                 .documentSplitter(splitter)
                 .embeddingStore(embeddingStore)
                 .embeddingModel(embeddingModel)
                 .build();
 
-        build.ingest(parentDoc);
+        ingestor.ingest(parentDoc);
 
         // Query and validate results
         final String retrieveQuery = "fundamental theory";
@@ -90,7 +90,7 @@ public class Neo4jEmbeddingStoreIngestorTest extends Neo4jEmbeddingStoreIngestor
         int maxSegmentSize = 250;
         DocumentSplitter splitter = new DocumentByRegexSplitter(expectedQuery, expectedQuery, maxSegmentSize, 0);
 
-        final Neo4jEmbeddingStoreIngestor build = Neo4jEmbeddingStoreIngestor.builder()
+        final Neo4jEmbeddingStoreIngestor ingestor = Neo4jEmbeddingStoreIngestor.builder()
                 .documentSplitter(splitter)
                 .embeddingStore(neo4jEmbeddingStore)
                 .embeddingModel(embeddingModel)
@@ -100,7 +100,7 @@ public class Neo4jEmbeddingStoreIngestorTest extends Neo4jEmbeddingStoreIngestor
                 .userPrompt("mock prompt user")
                 .systemPrompt("mock prompt system")
                 .build();
-        build.ingest(parentDoc);
+        ingestor.ingest(parentDoc);
         final String retrieveQuery = "naruto";
         List<Content> results = retriever.retrieve(Query.from(retrieveQuery));
         commonResults(results, retrieveQuery);
@@ -136,7 +136,7 @@ public class Neo4jEmbeddingStoreIngestorTest extends Neo4jEmbeddingStoreIngestor
         final String expectedQueryChild = "\\. ";
         DocumentSplitter childSplitter = new DocumentByRegexSplitter(expectedQueryChild, expectedQuery, 150, 0);
 
-        final Neo4jEmbeddingStoreIngestor build = Neo4jEmbeddingStoreIngestor.builder()
+        final Neo4jEmbeddingStoreIngestor ingestor = Neo4jEmbeddingStoreIngestor.builder()
                 .documentSplitter(parentSplitter)
                 .documentChildSplitter(childSplitter)
                 .driver(driver)
@@ -145,7 +145,7 @@ public class Neo4jEmbeddingStoreIngestorTest extends Neo4jEmbeddingStoreIngestor
                 .embeddingModel(embeddingModel)
                 .build();
         // Index the document into Neo4j as parent-child nodes
-        build.ingest(doc);
+        ingestor.ingest(doc);
 
         final String retrieveQuery = "Machine Learning";
         List<Content> results = retriever.retrieve(Query.from(retrieveQuery));
@@ -192,7 +192,7 @@ public class Neo4jEmbeddingStoreIngestorTest extends Neo4jEmbeddingStoreIngestor
         DocumentSplitter childSplitter =
                 new DocumentByRegexSplitter(expectedQueryChild, expectedQuery, maxSegmentSize, 0);
 
-        final Neo4jEmbeddingStoreIngestor build = Neo4jEmbeddingStoreIngestor.builder()
+        final Neo4jEmbeddingStoreIngestor ingestor = Neo4jEmbeddingStoreIngestor.builder()
                 .documentSplitter(parentSplitter)
                 .documentChildSplitter(childSplitter)
                 .driver(driver)
@@ -203,7 +203,7 @@ public class Neo4jEmbeddingStoreIngestorTest extends Neo4jEmbeddingStoreIngestor
                 .embeddingModel(embeddingModel)
                 .build();
         // Index the document into Neo4j as parent-child nodes
-        build.ingest(doc);
+        ingestor.ingest(doc);
 
         final String retrieveQuery = "Machine Learning";
         List<Content> results = retriever.retrieve(Query.from(retrieveQuery));
--- a/langchain4j-community-core/src/test/java/dev/langchain4j/community/embedding/ParentChildEmbeddingStoreIngestorTest.java
+++ b/langchain4j-community-core/src/test/java/dev/langchain4j/community/embedding/ParentChildEmbeddingStoreIngestorTest.java
@@ -10,6 +10,7 @@ import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.verifyNoMoreInteractions;
 import static org.mockito.Mockito.when;
 
+import dev.langchain4j.community.store.embedding.ParentChildEmbeddingStoreIngestor;
 import dev.langchain4j.data.document.Document;
 import dev.langchain4j.data.document.DocumentSplitter;
 import dev.langchain4j.data.document.DocumentTransformer;
@@ -109,7 +110,7 @@ public class ParentChildEmbeddingStoreIngestorTest {
                 asList(Embedding.from(new float[] {4}), Embedding.from(new float[] {7}));
         final List<Embedding> sixthEmbeddings =
                 asList(Embedding.from(new float[] {5}), Embedding.from(new float[] {6}));
-        
+
         when(embeddingModel.embedAll(any())).thenAnswer(invocation -> {
             List<TextSegment> segments = invocation.getArgument(0);
 


content-retrievers/langchain4j-community-neo4j-retriever/pom.xml
```diff
@@ -68,6 +68,18 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-community-neo4j</artifactId>
+            <version>1.0.0-beta3-SNAPSHOT</version>
```
Martin7-1, line 74:
```suggestion
            <version>${project.version}</version>
```

content-retrievers/langchain4j-community-neo4j-retriever/pom.xml
```diff
@@ -68,6 +68,18 @@
...
+            <scope>compile</scope>
+        </dependency>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j</artifactId>
+            <version>1.0.0-beta3-SNAPSHOT</version>
```
Martin7-1, line 80:
```suggestion
            <version>${project.version}</version>
```

content-retrievers/langchain4j-community-neo4j-retriever/src/test/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4JText2CypherRetrieverIT.java
```diff
@@ -10,10 +10,18 @@
 import java.util.List;
 import org.junit.jupiter.api.Test;
 import org.junit.jupiter.api.condition.EnabledIfEnvironmentVariable;
+import org.neo4j.driver.Session;
 
 @EnabledIfEnvironmentVariable(named = "OPENAI_API_KEY", matches = ".+")
-class Neo4JText2CypherRetrieverIT extends Neo4jText2CypherRetrieverBaseTest {
+class Neo4JText2CypherRetrieverIT extends Neo4jRetrieverBaseTest {
```
Martin7-1, line 16:
```suggestion
class Neo4jText2CypherRetrieverIT extends Neo4jRetrieverBaseTest {
```

content-retrievers/langchain4j-community-neo4j-retriever/src/test/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4JText2CypherRetrieverTest.java
```diff
@@ -13,12 +13,20 @@
 import org.junit.jupiter.api.extension.ExtendWith;
 import org.mockito.Mock;
 import org.mockito.junit.jupiter.MockitoExtension;
+import org.neo4j.driver.Session;
 
 @ExtendWith(MockitoExtension.class)
-public class Neo4JText2CypherRetrieverTest extends Neo4jText2CypherRetrieverBaseTest {
+public class Neo4JText2CypherRetrieverTest extends Neo4jRetrieverBaseTest {
```
Martin7-1, line 19:
```suggestion
public class Neo4jText2CypherRetrieverTest extends Neo4jRetrieverBaseTest {
```

content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4jEmbeddingRetriever.java
```diff
@@ -0,0 +1,407 @@
...
+
+    private <T> List<T> getList(EmbeddingSearchRequest request, Function<EmbeddingMatch<TextSegment>, T> embeddingMapFunction) {
+        return embeddingStore.search(request)
+                .matches()
+                .stream()
+                .map(embeddingMapFunction)
+                .toList();
+    }
+
+    public static class Builder<T extends Builder> {
```
Martin7-1, line 264: I thinks this builder does not need to be generic type. WDYT?

content-retrievers/langchain4j-community-neo4j-retriever/pom.xml
```diff
@@ -68,6 +68,18 @@
             <version>${project.version}</version>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-community-neo4j</artifactId>
+            <version>1.0.0-beta3-SNAPSHOT</version>
+            <scope>compile</scope>
```
Martin7-1, line 75: nit: compile is the deafult scope.

Martin7-1 (commented):
@vga91 Thank you!

content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4jEmbeddingRetriever.java
vga91: Yeah sorry, it would have been better to write it on the PR description, 
however the idea is to extend it (in a subsequent PR) by creating an ad-hoc parent-child retriever and other retriever (like [here](https://neo4j.com/labs/genai-ecosystem/langchain/#_advanced_rag_strategies) in langchain python) 
[in this way](https://github.com/vga91/langchain4j-community/commit/785f4b3627797f5c7eae8e77867803665eada39b#diff-a365e054db9a04b0a842db95f2e979c6c4a34e4d34026be5eea2b5a811bc8032R9-R12), 
as required in the issue #111 and as done in langchain python [in a similar way](https://python.langchain.com/api_reference/langchain/retrievers/langchain.retrievers.parent_document_retriever.ParentDocumentRetriever.html).

So I would leave this Neo4jEmbeddingRetriever as a fully-customizable GraphRAG retriever, like using the ones present [here](https://graphrag.com/reference/graphrag/basic-retriever/).

But of course, happy to change it if it can be done in a better way :D

Martin7-1: Thanks for your explanation, I got it now :)

Martin7-1 (commented):
@vga91 Thank you! All changes look good to me.

But I'm wondering if `Neo4jEmbeddingRetriever` is really a `ContentRetriever`. According to your implementation, `Neo4jEmbeddingRetriever` answers via llm and retrieved context, which looks like it's already a demo workflow of a rag system, not just the a `ContentRetriever`, according to the javadoc, the `ContentRetriever` should only retrieve contents from an underlying data source using a given Query.

vga91:
> @vga91 Thank you! All changes look good to me.
> 
> But I'm wondering if `Neo4jEmbeddingRetriever` is really a `ContentRetriever`. According to your implementation, `Neo4jEmbeddingRetriever` answers via llm and retrieved context, which looks like it's already a demo workflow of a rag system, not just the a `ContentRetriever`, according to the javadoc, the `ContentRetriever` should only retrieve contents from an underlying data source using a given Query.

Yeah it makes sense.
Maybe we can call it similarly to what we did in [langchain for python](https://github.com/neo4j-examples/rag-demo/blob/main/rag_demo/vector_chain.py#L93), 
calling it e.g. `RetrievalQAChain` and changing the `retrieve()` method to `invoke()`?

Martin7-1:
> Yeah it makes sense. Maybe we can call it similarly to what we did in [langchain for python](https://github.com/neo4j-examples/rag-demo/blob/main/rag_demo/vector_chain.py#L93), calling it e.g. `RetrievalQAChain` and changing the `retrieve()` method to `invoke()`?

@vga91 I think it's good, you can directly implement `Chain` interface which is in `dev.langchain4j.chain` package. 

BTW, can we decouple it from Neo4j? It seems that `RetrievalQAChain` only require the LLM and `ContentRetriever`, and a generic `RetrievalQAChain` through interface (i.e. `ChatModel`/`StreamingChatModel` and `ContentRetriever`) would be a better choice. Users can define their own prompt and pass it as constructor's parameter. And I think put it in `langchain4j-community-core` is a better choice.

In addition, we could put the use of Neo4j as `ContentRetriever` as an example elsewhere (e.g. a new `/examples` folder in the repository)

WDYT?

vga91:
@Martin7-1 Hi, I created a `RetrievalQAChain` [here](https://github.com/langchain4j/langchain4j-community/pull/147/commits/2191d4eb129ef7f017fee458af5b8d6317a18c47#diff-01ba7dd4c792aab684dab144fb07e6595cc2b356d6836b21120d391699154cdbR17) and removed the promptAnswer stuff from [here](https://github.com/langchain4j/langchain4j-community/pull/147/commits/2191d4eb129ef7f017fee458af5b8d6317a18c47#diff-8173d660916b8b64a573577226dd73e9255ae8347483743664021408002d8c63R226).
At the moment I created 2 tests `Neo4jEmbeddingRetrieverQAChainIT` and `Neo4jEmbeddingRetrieverQAChainTest` with the Neo4j Retriever used in the RetrievialQAChain.
Can you check if it's ok?

Maybe we can also add examples [here](https://github.com/langchain4j/langchain4j-examples/tree/main/neo4j-example)?

Martin7-1:
@vga91 Thank you! Sorry for the late reply. Will review it this weekend :)

> Maybe we can also add examples [here](https://github.com/langchain4j/langchain4j-examples/tree/main/neo4j-example)?

I think it's good, just go ahead!

langchain4j-community-core/src/main/java/dev/langchain4j/community/rag/content/RetrievalQAChain.java
```diff
@@ -0,0 +1,94 @@
...
+import dev.langchain4j.rag.AugmentationRequest;
+import dev.langchain4j.rag.AugmentationResult;
+import dev.langchain4j.rag.DefaultRetrievalAugmentor;
+import dev.langchain4j.rag.RetrievalAugmentor;
+import dev.langchain4j.rag.content.injector.DefaultContentInjector;
+import dev.langchain4j.rag.content.retriever.ContentRetriever;
+import dev.langchain4j.rag.query.Metadata;
+import dev.langchain4j.rag.query.Query;
+
+public class RetrievalQAChain implements Chain<Query, String> {
```
Martin7-1, line 17: Could you please add javadocs about its usage?

Also, what about moving it to `dev.langchain4j.community.chain` package?

embedding-stores/langchain4j-community-neo4j/src/main/java/dev/langchain4j/community/store/embedding/neo4j/Neo4jEmbeddingStore.java
```diff
@@ -189,6 +195,7 @@ public Neo4jEmbeddingStore(
         this.metadataPrefix = getOrDefault(metadataPrefix, "");
         this.textProperty = getOrDefault(textProperty, DEFAULT_TEXT_PROP);
         this.awaitIndexTimeout = getOrDefault(awaitIndexTimeout, DEFAULT_AWAIT_INDEX_TIMEOUT);
+        this.additionalParams = getOrDefault(additionalParams, Map.of());
```
Martin7-1, line 198:
```suggestion
        this.additionalParams = copy(additionalParams);
```

nit: just to keep immutable.

content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4jEmbeddingRetriever.java
```diff
@@ -0,0 +1,376 @@
...
+import dev.langchain4j.store.embedding.EmbeddingMatch;
+import dev.langchain4j.store.embedding.EmbeddingSearchRequest;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.function.Function;
+import org.neo4j.driver.Driver;
+import org.neo4j.driver.Session;
+
+public class Neo4jEmbeddingRetriever implements ContentRetriever {
```
Martin7-1, line 32: Should we still need this `Neo4jEmbeddingRetriever` as `ContentRetriever`? As I mentioned before, the functionality of `Neo4jEmbeddingRetriever` is not match `ContentRetriever`. So I think we can remove it.

BTW, we could just use `EmbeddingStoreContentRetriever` as `ContentRetriever` of `Neo4j`.

WDYT?

langchain4j-community-core/src/main/java/dev/langchain4j/community/rag/content/RetrievalQAChain.java
```diff
@@ -0,0 +1,94 @@
...
+        }
+
+        public Builder retrievalAugmentor(RetrievalAugmentor retrievalAugmentor) {
+            this.retrievalAugmentor = retrievalAugmentor;
+            return this;
+        }
+
+        public RetrievalQAChain build() {
+            if (retrievalAugmentor == null) {
+                try {
```
Martin7-1, line 80: I do not understand why we need this `try-catch`...

Looks like it's all good to just keep it simple and throw `queryRouter cannot be null`.

Martin7-1 (commented):
@vga91 Thank you!

content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4jEmbeddingRetriever.java
vga91: Yeah indeed the retrieve() method of the `Neo4jEmbeddingRetriever` can be replaced with the `EmbeddingStoreContentRetriever`.

But I wonder if it might be worth putting what is in the [index(..)](https://github.com/langchain4j/langchain4j-community/pull/147/files#diff-8173d660916b8b64a573577226dd73e9255ae8347483743664021408002d8c63R123) and [the builder](https://github.com/langchain4j/langchain4j-community/pull/147/files#diff-8173d660916b8b64a573577226dd73e9255ae8347483743664021408002d8c63R359) just as examples, instead of making a separate class too.

Because there is an [EmbeddingStoreIngestor](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-core/src/main/java/dev/langchain4j/store/embedding/EmbeddingStoreIngestor.java) available, but it saves embeddings starting from a Document and a splitter.

While the index() method, also because the intention is to extend it [as here](https://github.com/langchain4j/langchain4j-community/pull/147#discussion_r2066016416), also save the result in neo4j nodes/relationships and perform prompts (to easily implement eventually e.g. the [hypothetical retriever](https://graphrag.com/reference/graphrag/hypothetical-question-retriever/)),
and a childSplitter (to implement e.g. a [parent-child retriever](https://graphrag.com/reference/graphrag/parent-child-retriever/)).

Perhaps we could create a `Neo4jEmbeddingStoreIngestor` that extends `EmbeddingStoreIngestor`?
Or even, it seems feasible, a `ParentChildEmbeddingStoreIngestor` that extends `EmbeddingStoreIngestor` (with a childSplitter) and/or a `Neo4jEmbeddingStoreIngestor` extends `ParentChildEmbeddingStoreIngestor`? 
Wdyt?

Martin7-1: > But I wonder if it might be worth putting what is in the [index(..)](https://github.com/langchain4j/langchain4j-community/pull/147/files#diff-8173d660916b8b64a573577226dd73e9255ae8347483743664021408002d8c63R123) and [the builder](https://github.com/langchain4j/langchain4j-community/pull/147/files#diff-8173d660916b8b64a573577226dd73e9255ae8347483743664021408002d8c63R359) just as examples, instead of making a separate class too.

From my point of view, I don't think the index method is the responsibility of ContentRetriever, and involving model-related calls here would give the user less freedom, and placing it in examples would be a better choice. Our intention is just give users components and let them assemble it freely.

> Perhaps we could create a Neo4jEmbeddingStoreIngestor that extends EmbeddingStoreIngestor?
Or even, it seems feasible, a ParentChildEmbeddingStoreIngestor that extends EmbeddingStoreIngestor (with a childSplitter) and/or a Neo4jEmbeddingStoreIngestor extends ParentChildEmbeddingStoreIngestor?
Wdyt?

Hmm...I think it's a good idea, and we could put `ParentChildEmbeddingStoreIngestor` into `langchain4j-community-core` maybe.

langchain4j-community-core/src/main/java/dev/langchain4j/community/rag/content/RetrievalQAChain.java
vga91: Indeed it's fine even so. Changed

content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/Neo4jEmbeddingRetriever.java
vga91: Ok, added ParentChildEmbeddingStoreIngestor and Neo4jEmbeddingStoreIngestor

Martin7-1 (approved):
@vga91 Thank you!

BTW, could you please add some tests about `ParentChildEmbeddingStoreIngestor`?

Martin7-1:
> @vga91 Thank you!

BTW, could you please add some tests about `ParentChildEmbeddingStoreIngestor`?

Martin7-1 (approved):
@vga91 Thank you!

PR #151, opened 2025-04-24
[DRAFT] Fix Neo4jEmbedding ensureStoreIsEmpty test

vga91:
## Change

- Fix flaky Neo4jEmbedding ensureStoreIsEmpty test.

See here: https://github.com/langchain4j/langchain4j-community/actions/runs/14649577148/job/41111886216?pr=116#step:5:5340

Theoretically the `DROP INDEX` after the tests is not strictly needed, since the tests work anyway because there is an `IF NOT EXISTS` that does not cause the execution to fail if it finds an equivalent index.

Not sure why locally it always works even when starting all the tests together and why on the IC it seems to always fail to show up and also fails on the PR where Neo4jEmbeddingStore was not changed.

But probably the tests are run in different order and it could happen that `EmbeddingStore.search(..)` method which calls the `queryNodes(indexName, ...)` is done after the `DROP INDEX ...` but before the `CREATE VECTOR INDEX ...` of the `Neo4jEmbeddingStore.builder()...build()`.

---

## General checklist


Martin7-1 (approved):
@vga91 Thank you!

Issue #146, opened 2025-04-23
[BUG] QwenHelper.messageAccumulator() java.util.NoSuchElementException


miracle-bean:
## Describe the bug

![Image](https://github.com/user-attachments/assets/c801b9ed-90aa-4420-982c-7e06983277ee)
acc.removeLast() make it not work

## Log and Stack trace
java.util.NoSuchElementException: null
	at java.base/java.util.LinkedList.getLast(LinkedList.java:268)
	at dev.langchain4j.community.model.dashscope.QwenHelper.lambda$messageAccumulator$26(QwenHelper.java:454)
	at java.base/java.util.stream.ReduceOps$1ReducingSink.accept(ReduceOps.java:80)
	at java.base/java.util.ArrayList$ArrayListSpliterator.forEachRemaining(ArrayList.java:1708)
	at java.base/java.util.stream.AbstractPipeline.copyInto(AbstractPipeline.java:509)
	at java.base/java.util.stream.AbstractPipeline.wrapAndCopyInto(AbstractPipeline.java:499)
	at java.base/java.util.stream.ReduceOps$ReduceOp.evaluateSequential(ReduceOps.java:921)
	at java.base/java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.base/java.util.stream.ReferencePipeline.reduce(ReferencePipeline.java:667)
	at dev.langchain4j.community.model.dashscope.QwenHelper.sanitizeMessages(QwenHelper.java:416)
	at dev.langchain4j.community.model.dashscope.QwenHelper.toQwenMessages(QwenHelper.java:80)
	at dev.langchain4j.community.model.dashscope.QwenHelper.toGenerationParam(QwenHelper.java:617)
	at dev.langchain4j.community.model.dashscope.QwenStreamingChatModel.generateByNonMultimodalModel(QwenStreamingChatModel.java:138)
	at dev.langchain4j.community.model.dashscope.QwenStreamingChatModel.doChat(QwenStreamingChatModel.java:205)
	at dev.langchain4j.model.chat.StreamingChatLanguageModel.chat(StreamingChatLanguageModel.java:68)
	at dev.langchain4j.service.AiServiceTokenStream.start(AiServiceTokenStream.java:122)

## Please complete the following information:
- LangChain4j Community version: 1.0.0-beta3
- LangChain4j Community Component you used (e.g. `langchain4j-community-clickhouse`): 1.0.0-beta3
- Java version: 17
[...]

PR #154, opened 2025-04-25
DashScope: Support the 'dimension' parameter in QwenEmbeddingModel, and more.

### LINKED ISSUE #105: [FEATURE] dashscope-embedding-model dimensions
There is no place to set dimensions

jiangsier-xyz:
## Issue

Closes #105 #146

## Change

- Support the 'dimension' parameter in QwenEmbeddingModel
- Add the fields isMultimodalModel and supportIncrementalOutput to QwenChatRequestParameters
  - Whether a model is multimodal and whether it supports incremental output is related to the model name. The model name is set in ChatRequestParameters and can change with each call. Therefore, isMultimodalModel and supportIncrementalOutput should also be allowed to be set at each call without needing to recreate the ChatModel instance.
  - Regarding supportIncrementalOutput: some models do not support incremental output (e.g., Qwen-MT), while some models only support incremental output (e.g., [QwQ](https://www.alibabacloud.com/help/en/model-studio/qwq#62c72012bc2sw)). Specifying this parameter helps the QwenStreamingChatModel to **standardize on incremental output**.
- Retain as much response information as possible in stream mode
  - In stream mode, merge all fields from the result each time, rather than only retaining the content that langchain4j is concerned with. The complete response information can be quite useful for applications, such as debugging purposes.
- Allow modification of the system message during chat, but does not discard historical messages. And fix issues.
- To accelerate testing speed and reduce costs, reduce the number of models
  - Only newer and stable models are being tested.

## General checklist

- [X] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)


Martin7-1:
@jiangsier-xyz Thank you! There are some failed build on javadoc, could you please check?

Martin7-1 (approved):
@jiangsier-xyz Thank you!

PR #171, opened 2025-04-30
redis: RedisChatMemoryStore  ttl & prefix support

kingster:
## Issue

Redis based RedisChatMemoryStore doesn't support ttl & prefix which is supported in the [python version](https://github.com/langchain-ai/langchain-redis/blob/main/libs/redis/langchain_redis/chat_message_history.py).

## Change
Added support for ttl & prefixKey in RedisChatMemoryStore.

## General checklist

- [x] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j


Martin7-1:
@kingster Thank you! Could you please test whether ttl works or not?

Martin7-1 (approved):
@kingster Thank you! All changes look good. But it needs some improvement on javadocs and tests:

1. Could you please add javadoc on the constructor/builder? Maybe explain: the format of keyPrefix (such as it should end with `:` or not) and the unit of `ttl` (second or millseconds?)
2. Maybe we also need to test whether `prefix` works or not.

kingster:
Hi @Martin7-1 

Updated the java docs & added another test_case with prefix.

Martin7-1 (approved):
@kingster Thank you!

PR #217, opened 2025-05-28
Add the ability to store the KnowledgeGraphWriter results in vector indexes via Neo4jEmbeddingStore

### LINKED ISSUE #216: [FEATURE] Add the ability to store the `KnowledgeGraphWriter ` results in vector indexes via Neo4jEmbeddingStore
## Context

See here: https://github.com/langchain4j/langchain4j-community/issues/185

Currently in Langchain4j we can create a Neo4j KnowledgeGraph via the [KnowledgeGraphWriter](https://github.com/langchain4j/langchain4j-community/blob/main/content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/KnowledgeGraphWriter.java) class, 
but to retrieve the results we would have to use manual queries (e.g. `MATCH (c:Chunk)-[]->(:Document) RETURN c).

No vectors are saved, instead just string texts are saved.

## Use case solution

To create something similar, i.e. with text embedded from the document texts,
we could add an argument to `KnowledgeGraphWriter` builder, for example:
```
String retrievalQuery = """
        MATCH (chunk)-[:PART_OF]->(d:Document)
        // aggregate chunk-details
        WITH d, collect(DISTINCT {chunk: chunk, score: score}) AS chunks, avg(score) as avg_score
        // ....
        """;

final Neo4jEmbeddingStore neo4jEmbeddingStore = Neo4jEmbeddingStore.builder()
        .driver(null)
        .dimension(384)
        .label("Chunk")
        .retrievalQuery(retrievalQuery).build();

knowledgeGraphWriter = KnowledgeGraphWriter.builder()
  .embeddingStore(embeddingStore)
  // .... other methods
  .build()
```

so it can save vectors in nodes with label `Chunk` to be retrieved in this way
```
List<EmbeddingMatch<TextSegment>> matches = neo4jEmbeddingStore.search(null).matches();

/* 
which will search `Chunk` with embedding using a query like:
CALL db.index.vector.queryNodes($indexName, $maxResults, $embeddingValue) YIELD node, score WHERE score >= $minScore
MATCH (node)-[:PART_OF]->(d:Document) 
*/
```

\cc @Martin7-1

vga91:
## Issue

Closes #216

## Change

Currently in Langchain4j we can create a Neo4j KnowledgeGraph via the [KnowledgeGraphWriter](https://github.com/langchain4j/langchain4j-community/blob/main/content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/KnowledgeGraphWriter.java) class,
but to retrieve the results we would have to use manual queries (e.g. `MATCH (c:Chunk)-[]->(:Document) RETURN c).

To create something similar, i.e. with text embedded from the document texts,
added an argument  to `KnowledgeGraphWriter` constructor, so that we can store and retrieve vector via Neo4jEmbeddingStore.

## General checklist

### DIFF
--- a/content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/KnowledgeGraphWriter.java
+++ b/content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/KnowledgeGraphWriter.java
@@ -7,7 +7,14 @@ import static dev.langchain4j.internal.ValidationUtils.ensureNotNull;
 
 import dev.langchain4j.Experimental;
 import dev.langchain4j.community.data.document.graph.GraphDocument;
+import dev.langchain4j.community.data.document.graph.GraphNode;
+import dev.langchain4j.community.store.embedding.neo4j.Neo4jEmbeddingStore;
 import dev.langchain4j.data.document.Document;
+import dev.langchain4j.data.document.Metadata;
+import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.data.segment.TextSegment;
+import dev.langchain4j.model.embedding.EmbeddingModel;
+import java.util.ArrayList;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
@@ -34,6 +41,8 @@ public class KnowledgeGraphWriter {
     final String sanitizedTextProperty;
 
     private final Neo4jGraph graph;
+    private final Neo4jEmbeddingStore embeddingStore;
+    private EmbeddingModel embeddingModel = null;
 
     public KnowledgeGraphWriter(
             Neo4jGraph graph,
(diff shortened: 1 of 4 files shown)


content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/KnowledgeGraphWriter.java
```diff
@@ -34,23 +41,38 @@ public class KnowledgeGraphWriter {
...
             String textProperty,
             String relType,
-            String constraintName) {
+            String constraintName,
+            Neo4jEmbeddingStore embeddingStore,
+            EmbeddingModel embeddingModel) {
         this.graph = ensureNotNull(graph, "graph");
+
+        this.embeddingStore = embeddingStore;
+        final boolean storeIsNull = this.embeddingStore == null;
```
Martin7-1, line 59: I think `storeIsNull` will always be true as `this.embeddingStore` will always be `null` before calling ctor.

Martin7-1 (commented):
@vga91 Thank you!

content-retrievers/langchain4j-community-neo4j-retriever/src/main/java/dev/langchain4j/community/rag/content/retriever/neo4j/KnowledgeGraphWriter.java
vga91: It works correctly since there is the `this.embeddingStore = embeddingStore;` right before, 
but indeed if you want I can put `final boolean storeIsNull = embeddingStore == null;` which is perhaps safer as it would also work without the `this.embeddingStore = embeddingStore;`?

Martin7-1: Oh, sorry I missed it. Don't worry, everything is good now :)

Martin7-1 (approved):
@vga91 Thank you!

PR #243, opened 2025-06-15
DashScope: Introduce enableSanitizeMessages option

jiangsier-xyz:
## Change

1. Make sanitizeMessages() throw more appropriate exception types.
2. Qwen includes a lot models. These models may not follow a uniform message order constraint. Therefore, an enableSanitizeMessages option has been introduced to control whether the langchain4j framework sanitize input messages.
3. I tested the latest qwen-max (which should be a qwen3), and the constraint on message order seems to have become more insensitive.   
For instance, messages like ("assistant", "assistant", "system") do not trigger any error. This differs from the description in the documentation, which says that the system message must be placed first in the list. The actual requirements for message order have become unclear.   
However, the qwen2 model remains sensitive to message order. Considering that people use the models through proxy names (such as qwen-max), the default value for the enableSanitizeMessages option has been set to false. Which might be a breaking change.

## General checklist

### DIFF
--- a/models/langchain4j-community-dashscope/src/main/java/dev/langchain4j/community/model/dashscope/QwenChatModel.java
+++ b/models/langchain4j-community-dashscope/src/main/java/dev/langchain4j/community/model/dashscope/QwenChatModel.java
@@ -112,6 +112,7 @@ public class QwenChatModel implements ChatModel {
                 .supportIncrementalOutput(qwenParameters.supportIncrementalOutput())
                 .enableThinking(qwenParameters.enableThinking())
                 .thinkingBudget(qwenParameters.thinkingBudget())
+                .enableSanitizeMessages(qwenParameters.enableSanitizeMessages())
                 .custom(copyIfNotNull(qwenParameters.custom()))
                 .build();
 
--- a/models/langchain4j-community-dashscope/src/main/java/dev/langchain4j/community/model/dashscope/QwenChatRequestParameters.java
+++ b/models/langchain4j-community-dashscope/src/main/java/dev/langchain4j/community/model/dashscope/QwenChatRequestParameters.java
@@ -66,7 +66,7 @@ public class QwenChatRequestParameters extends DefaultChatRequestParameters {
     private final Integer thinkingBudget;
     /**
      * Specifies whether to sanitize messages before sending to llm provider.
-     * Default value is false.
+     * Default value is true.
      */
     private final Boolean enableSanitizeMessages;
     /**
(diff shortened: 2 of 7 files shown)


Martin7-1 (approved):
@jiangsier-xyz Thank you!

Martin7-1:
> I tested the latest qwen-max (which should be a qwen3), and the constraint on message order seems to have become more insensitive.
For instance, messages like ("assistant", "assistant", "system") do not trigger any error. This differs from the description in the documentation, which says that the system message must be placed first in the list. The actual requirements for message order have become unclear.
However, the qwen2 model remains sensitive to message order. Considering that people use the models through proxy names (such as qwen-max), the default value for the enableSanitizeMessages option has been set to false. Which might be a breaking change.

I think we should keep the order to keep compatibility, although the new models become insensitive to the message order. BTW, current implementation on `Qwen(Streaming)ChatModel` is still keep the order right?

jiangsier-xyz:
> > I tested the latest qwen-max (which should be a qwen3), and the constraint on message order seems to have become more insensitive.
> > For instance, messages like ("assistant", "assistant", "system") do not trigger any error. This differs from the description in the documentation, which says that the system message must be placed first in the list. The actual requirements for message order have become unclear.
> > However, the qwen2 model remains sensitive to message order. Considering that people use the models through proxy names (such as qwen-max), the default value for the enableSanitizeMessages option has been set to false. Which might be a breaking change.
> 
> I think we should keep the order to keep compatibility, although the new models become insensitive to the message order. 

It's OK for me to let enableSanitizeMessages default value to be true.

> BTW, current implementation on `Qwen(Streaming)ChatModel` is still keep the order right?

Yes.

jiangsier-xyz:
> > > I tested the latest qwen-max (which should be a qwen3), and the constraint on message order seems to have become more insensitive.
> > > For instance, messages like ("assistant", "assistant", "system") do not trigger any error. This differs from the description in the documentation, which says that the system message must be placed first in the list. The actual requirements for message order have become unclear.
> > > However, the qwen2 model remains sensitive to message order. Considering that people use the models through proxy names (such as qwen-max), the default value for the enableSanitizeMessages option has been set to false. Which might be a breaking change.
> > 
> > 
> > I think we should keep the order to keep compatibility, although the new models become insensitive to the message order.
> 
> It's OK for me to let enableSanitizeMessages default value to be true.
> 
> > BTW, current implementation on `Qwen(Streaming)ChatModel` is still keep the order right?
> 
> Yes.

Btw, after setting enableSanitizeMessages default value to be true, the PR is not a breaking change anymore.

Martin7-1:
> Btw, after setting enableSanitizeMessages default value to be true, the PR is not a breaking change anymore.

OK. I got it. I will merge it for now.

BTW, just like [langchain4j-anthropic](https://github.com/langchain4j/langchain4j/pull/3221), maybe we should remove sanitize messages in the future. (Maybe when the official docs say the order does not matter anymore).

I am curious about qwen3's output with `enableSanitizeMessages` enabled and disabled. Maybe it need some explanation and comparison in the docs.

PR #271, opened 2025-06-27
fix ChatResponseMetadata type not inconsistent

### LINKED ISSUE #270: [BUG] When using the streaming output of aiService and streamingChatModel, the metadata types in ChatResponse are inconsistent
The metadata type in the ChatResponse object of the onCompleteResponse method of the aiService's streaming has been fixed as ChatResponseMetadata.

The following is the basic metadata of the streamingChatModel during streaming chat, which can be seen as the metadata corresponding to the model.

Image
When I switched to using the streaming chat function of aiService, I found that metadata belongs to the top-level ChatResponseMetadata class, rather than the Metadata class corresponding to the model.

Image

lzx1026078111:
## Issue

Closes #270 

## Change

1. QwenChatResponseMetadata override the `toBuilder()` function.
2. reviewed other models, only the model which extend the ChatResponseMetadata may have this problem. In langchain-community, only Qwen extend the ChatResponseMetadata, so no other model have this problem.

## General checklist


jiangsier-xyz:
LGTM.

PR #326, opened 2025-08-05
Add HybridEmbeddingStore: In-Memory Embeddings with Disk-Based Chunk

harikrishna553:
### Summary

This PR introduces a new `MemFileEmbeddingStore<Embedded>` implementation for the LangChain4j framework.

### Key Features

- Stores embeddings (float vectors) in memory for fast similarity search
- Stores embedded content (e.g., `TextSegment`) as files on disk to reduce memory usage
- Supports:
  - Brute-force similarity search
  - Serialization to/from JSON and file
  - Optional LRU cache for frequently accessed chunks
- Designed for medium-scale use cases where memory efficiency is critical

### Why This Is Useful

This hybrid design balances performance and scalability:
- In-memory vector lookup ensures quick search
- Disk-backed content allows handling of large datasets without exceeding JVM heap limits

### DIFF
--- a/embedding-stores/langchain4j-community-memfile/src/main/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStore.java
+++ b/embedding-stores/langchain4j-community-memfile/src/main/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStore.java
@@ -43,7 +43,7 @@ import org.slf4j.LoggerFactory;
 /**
  * An {@link EmbeddingStore} implementation that stores embeddings in memory
  * while persisting associated embedded content (e.g.,
- * {@link dev.langchain4j.data.segment.TextSegment}) as JSON files in a local
+ * {@link dev.langchain4j.data.segment.TextSegment}) as files in a local
  * directory.
  *
  * <p>
@@ -51,7 +51,7 @@ import org.slf4j.LoggerFactory;
  *
  * <ul>
  * <li>Embeddings are kept fully in memory for fast similarity search.</li>
- * <li>Embedded content is serialized to JSON and stored as separate files on
+ * <li>Embedded content is serialized and stored as separate files on
  * disk, reducing memory footprint for large content.</li>
  * <li>Optional LRU cache keeps frequently accessed embedded content in
  * memory.</li>
@@ -67,8 +67,7 @@ import org.slf4j.LoggerFactory;
  * to the configured {@code chunkStorageDirectory}.</li>
  * <li>Content is reloaded from disk on demand and optionally cached for
  * reuse.</li>
- * <li>Chunk files are named after the embedding ID with a {@code .json}
- * extension.</li>
+ * <li>Chunk files are named after the embedding ID</li>
  * </ul>
  *
  * <p>
@@ -140,6 +139,29 @@ public class MemFileEmbeddingStore<Embedded> implements EmbeddingStore<Embedded>
         this.cacheSize = Math.max(0, cacheSize);
         this.chunkCache = cacheSize > 0 ? createLRUCache(cacheSize) : new ConcurrentHashMap<>();
         createChunkStorageDirectory();
+    }
+
+    private static Path createDefaultChunkDirectory() {
+        try {
+            return Files.createTempDirectory("memfile-embedding-store");
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to create default chunk storage directory", e);
+        }
+    }
+
+    private static <K, V> Map<K, V> createLRUCache(int maxSize) {
+        // Calculate capacity so that (capacity * loadFactor) >= maxSize
+        // This avoids an unnecessary resize before reaching maxSize entries.
+        int capacity = (int) Math.ceil(maxSize / 0.75f) + 1;
+
+        Map<K, V> lru = new LinkedHashMap<K, V>(capacity, 0.75f, true) {
+            @Override
+            protected boolean removeEldestEntry(Map.Entry<K, V> eldest) {
+                return size() > maxSize; // Evict eldest when size exceeds maxSize
+            }
+        };
+
+        return Collections.synchronizedMap(lru);
     }
 
     @Override
@@ -402,7 +424,7 @@ public class MemFileEmbeddingStore<Embedded> implements EmbeddingStore<Embedded>
      * }</pre>
      *
      * @param storeSerializationStrategy the strategy to use for serialization; must not be {@code null}
-     * @param filePath the path where the serialized data should be written; must not be {@code null}
+     * @param filePath                   the path where the serialized data should be written; must not be {@code null}
      * @throws RuntimeException if the file cannot be created, written to, or if serialization fails
      * @see #deserializeFromFile(StoreSerializationStrategy, Path)
      * @see #serialize(StoreSerializationStrategy)
@@ -439,8 +461,8 @@ public class MemFileEmbeddingStore<Embedded> implements EmbeddingStore<Embedded>
      * }</pre>
      *
      * @param storeSerializationStrategy the strategy to use for serialization; must not be {@code null}
-     * @param filePath the file path as a string where the serialized data should be written; must not be {@code null} or blank
-     * @throws RuntimeException if the file cannot be created, written to, or if serialization fails
+     * @param filePath                   the file path as a string where the serialized data should be written; must not be {@code null} or blank
+     * @throws RuntimeException         if the file cannot be created, written to, or if serialization fails
      * @throws IllegalArgumentException if the file path is invalid
      * @see #deserializeFromFile(StoreSerializationStrategy, String)
      * @see #serializeToFile(StoreSerializationStrategy, Path)
@@ -490,9 +512,9 @@ public class MemFileEmbeddingStore<Embedded> implements EmbeddingStore<Embedded>
      * }</pre>
      *
      * @param storeSerializationStrategy the strategy to use for deserialization; must not be {@code null}
-     * @param data the serialized string representation of the embedding store; must not be {@code null} or blank
+     * @param data                       the serialized string representation of the embedding store; must not be {@code null} or blank
      * @return a new {@code MemFileEmbeddingStore} instance restored from the serialized data
-     * @throws RuntimeException if deserialization fails due to invalid data format, I/O errors, or strategy-specific issues
+     * @throws RuntimeException         if deserialization fails due to invalid data format, I/O errors, or strategy-specific issues
      * @throws IllegalArgumentException if the serialized data is malformed or incompatible
      * @see #serialize(StoreSerializationStrategy)
      * @see #deserializeFromFile(StoreSerializationStrategy, Path)
@@ -542,9 +564,9 @@ public class MemFileEmbeddingStore<Embedded> implements EmbeddingStore<Embedded>
      * }</pre>
      *
      * @param storeSerializationStrategy the strategy to use for deserialization; must not be {@code null}
-     * @param filePath the path to the file containing serialized store data; must not be {@code null}
+     * @param filePath                   the path to the file containing serialized store data; must not be {@code null}
      * @return a new {@code MemFileEmbeddingStore} instance restored from the file data
-     * @throws RuntimeException if the file cannot be read, doesn't exist, or if deserialization fails
+     * @throws RuntimeException         if the file cannot be read, doesn't exist, or if deserialization fails
      * @throws IllegalArgumentException if the file contains malformed or incompatible data
      * @see #serializeToFile(StoreSerializationStrategy, Path)
      * @see #deserialize(StoreSerializationStrategy, String)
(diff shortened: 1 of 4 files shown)


embedding-stores/langchain4j-community-hybrid-store/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-community</artifactId>
+        <version>1.3.0-beta9-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-community-hybrid-store</artifactId>
```
Martin7-1: I'm not sure if we should call it this name...

embedding-stores/langchain4j-community-hybrid-store/pom.xml
```diff
@@ -0,0 +1,64 @@
...
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-core</artifactId>
+            <version>${langchain4j.core.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j</artifactId>
+            <version>${langchain4j.core.version}</version>
```
Martin7-1:
```suggestion
            <version>${langchain4j.version}</version>
```

embedding-stores/langchain4j-community-hybrid-store/src/main/java/dev/langchain4j/store/embedding/inmemory/HybridEmbeddingStore.java
```diff
@@ -0,0 +1,563 @@
+package dev.langchain4j.store.embedding.inmemory;
```
Martin7-1: Please move it to `dev.langchain4j.community` package.

Also, I'm not sure if we should call it with `inmemory` package...

embedding-stores/langchain4j-community-hybrid-store/src/test/java/dev/langchain4j/store/embedding/inmemory/HybridEmbeddingStoreIt.java
```diff
@@ -0,0 +1,764 @@
...
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.io.TempDir;
+
+class HybridEmbeddingStoreIt {
```
Martin7-1:
```suggestion
class HybridEmbeddingStoreIT {
```

Please extend it with `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`

harikrishna553: Addressed by extending EmbeddingStoreWithFilteringIT

embedding-stores/langchain4j-community-hybrid-store/src/main/java/dev/langchain4j/store/embedding/inmemory/HybridEmbeddingStore.java
harikrishna553: @Martin7-1 Thank you for your review.

- The HybridEmbeddingStore extends InMemoryEmbeddingStore and reuses some of its internal logic. To enable this, it requires access to the entries variable in InMemoryEmbeddingStore. One possible solution is to change the visibility of entries to protected. I’m also open to renaming the package if that helps. Please share your thoughts on this approach.

- There is an alternative version of the same implementation that does not extend InMemoryEmbeddingStore. You can take a look at it [here](https://github.com/langchain4j/langchain4j/blob/8260fd2482c48c0c23f42cf91b74a94f373aad91/langchain4j/src/main/java/dev/langchain4j/store/embedding/hybrid/HybridEmbeddingStore.java). With this approach, I can refactor the class to dev.langchain4j.community package.

Both approaches have their pros and cons. I’ve verified the functionality by extending the EmbeddingStoreWithFilteringIT class for both the solutions, and everything is working as expected.

embedding-stores/langchain4j-community-hybrid-store/pom.xml
harikrishna553: Addressed

embedding-stores/langchain4j-community-hybrid-store/pom.xml
harikrishna553: Thanks for highlighting that. I'm open to renaming — do you have any suggestions that might better capture the module's purpose or scope?

One option could be:
Module: langchain4j-community-file-backed-memory-store
Class: FileBackedEmbeddingStore

Regarding the new module, this was a suggestion from @dliubarskyi on a separate PR (https://github.com/langchain4j/langchain4j/pull/3293).
@dliubarskyi , could you please share your thoughts here as well?

Martin7-1: What about calling it `langchain4j-community-memfile` and the class called `MemFileEmbeddingStore`

embedding-stores/langchain4j-community-hybrid-store/src/main/java/dev/langchain4j/store/embedding/inmemory/HybridEmbeddingStore.java
Martin7-1: I think the second way would be better as we want to keep consistent package name

embedding-stores/langchain4j-community-hybrid-store/pom.xml
harikrishna553: Addressed

embedding-stores/langchain4j-community-hybrid-store/src/main/java/dev/langchain4j/store/embedding/inmemory/HybridEmbeddingStore.java
harikrishna553: Agreed.

embedding-stores/langchain4j-community-memfile/src/main/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStore.java
```diff
@@ -0,0 +1,585 @@
...
+    private static Path createDefaultChunkDirectory() {
+        try {
+            return Files.createTempDirectory("memfile-embedding-store");
+        } catch (IOException e) {
+            throw new RuntimeException("Failed to create default chunk storage directory", e);
+        }
+    }
+
+    private static <K, V> Map<K, V> createLRUCache(int maxSize) {
+        return new LinkedHashMap<K, V>(16, 0.75f, true) {
```
Martin7-1: I'm wondering why the `LinkedHashMap` is initialized with capacity 16 instead of `maxSize`.

Also, looks like it should be thread-safe. WDYT?

embedding-stores/langchain4j-community-memfile/src/main/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStore.java
```diff
@@ -0,0 +1,585 @@
...
+     * @return A new MemFileEmbeddingStore instance with the specified directory
+     */
+    public MemFileEmbeddingStore<Embedded> withChunkStorageDirectory(Path chunkStorageDirectory) {
+        return new MemFileEmbeddingStore<>(entries, chunkStorageDirectory, cacheSize);
+    }
+
+    /**
+     * Serializes the store to JSON, including the in-memory index and chunk file paths.
+     */
+    public String serializeToJson() {
```
Martin7-1: I think for the serialize design, we should provide a interface to let users customize serialize strategy, maybe to json, to binary, or just to raw text.

WDYT?

embedding-stores/langchain4j-community-memfile/src/main/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStore.java
```diff
@@ -0,0 +1,585 @@
...
+    private static <K, V> Map<K, V> createLRUCache(int maxSize) {
+        return new LinkedHashMap<K, V>(16, 0.75f, true) {
+            @Override
+            protected boolean removeEldestEntry(Map.Entry<K, V> eldest) {
+                return size() > maxSize;
+            }
+        };
+    }
+
+    static class Entry<Embedded> {
```
Martin7-1: If there is no special need, please remove the redundant '@JsonProperty' annotation

embedding-stores/langchain4j-community-memfile/src/test/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStoreIt.java
```diff
@@ -0,0 +1,39 @@
...
+import dev.langchain4j.data.segment.TextSegment;
+import dev.langchain4j.model.embedding.EmbeddingModel;
+import dev.langchain4j.model.embedding.onnx.allminilml6v2q.AllMiniLmL6V2QuantizedEmbeddingModel;
+import dev.langchain4j.store.embedding.EmbeddingStore;
+import dev.langchain4j.store.embedding.EmbeddingStoreWithFilteringIT;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.util.UUID;
+
+public class MemFileEmbeddingStoreIt extends EmbeddingStoreWithFilteringIT {
```
Martin7-1:
```suggestion
public class MemFileEmbeddingStoreIT extends EmbeddingStoreWithFilteringIT {
```

embedding-stores/langchain4j-community-memfile/src/main/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStore.java
harikrishna553: That’s a good catch. How about we set the initial capacity to (int) Math.ceil(maxSize / 0.75f) + 1 , so the map won’t resize before it reaches maxSize, while still keeping the load factor at 0.75 for good performance.

```
private static <K, V> Map<K, V> createLRUCache(int maxSize) {
    // Calculate capacity so that (capacity * loadFactor) >= maxSize
    // This avoids an unnecessary resize before reaching maxSize entries.
    int capacity = (int) Math.ceil(maxSize / 0.75f) + 1;

    return new LinkedHashMap<K, V>(capacity, 0.75f, true) {
        @Override
        protected boolean removeEldestEntry(Map.Entry<K, V> eldest) {
            return size() > maxSize; // Evict eldest when size exceeds maxSize
        }
    };
}

```

embedding-stores/langchain4j-community-memfile/src/main/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStore.java
harikrishna553: Agreed, making the serialization strategy pluggable via an interface sounds like a good design improvement.

embedding-stores/langchain4j-community-memfile/src/main/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStore.java
Martin7-1: Looks good to me :)

harikrishna553: @Martin7-1  I want to replace CopyOnWriteArrayList with ConcurrentLinkedQueue to store the entries. What do you think?

embedding-stores/langchain4j-community-memfile/src/main/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStore.java
harikrishna553: Addressed

embedding-stores/langchain4j-community-memfile/src/test/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStoreIt.java
harikrishna553: Addressed

embedding-stores/langchain4j-community-memfile/src/main/java/dev/langchain4j/community/store/embedding/memfile/MemFileEmbeddingStore.java
Martin7-1: > @Martin7-1 I want to replace CopyOnWriteArrayList with ConcurrentLinkedQueue to store the entries. What do you think?

I think this idea is good because `CopyOnWriteArrayLIst` can have the problem of copying too much if you write a lot at the same time.

Martin7-1 (approved):
@harikrishna553 Thank you! Great job!

harikrishna553:
@Martin7-1 Thank you for reviewing and merging the PR. Really appreciate your time in reviewing this.

PR #327, opened 2025-08-07
feat: Integrate MMR algorithm into ReRankingContentAggregator 

### LINKED ISSUE #325: [FEATURE] Integrate MMR into ReRankingContentAggregator
### Is your feature request related to a problem? Please describe.

The current `ReRankingContentAggregator` implementation is very effective at fusing content from multiple retrieval sources using Reciprocal Rank Fusion (RRF) and then re-ranking it based on a `ScoringModel`. However, the final re-ranking step relies solely on the relevance score provided by the `ScoringModel`, without explicitly considering the diversity of the selected documents.

This can lead to a few problems:
1.  **Redundancy:** The top-ranked documents might contain highly similar or redundant information. While these documents are all highly relevant to the user query, they do not provide a broad and comprehensive context for the LLM.
2.  **Limited Scope:** By not balancing relevance with diversity, the system might miss out on important, but slightly less relevant, information that covers a different aspect of the query. This can lead to the LLM generating a narrow or one-sided answer.
3.  **Suboptimal LLM Context:** The goal of RAG is to provide the best possible context to the LLM. A list of highly relevant but redundant documents is not as effective as a list of documents that are both relevant and diverse.

### Describe the solution you'd like

I would like to propose integrating the **Maximal Marginal Relevance (MMR)** algorithm into the `ReRankingContentAggregator`. This would allow the aggregator to re-rank the fused content not only by relevance but also by diversity, providing a more balanced and informative context to the LLM.

The proposed solution would be as follows:
1.  **Introduce an `MmrReranker` class**: A dedicated, low-level utility class `MmrReranker` would be created (or an existing one would be used if available in the codebase, e.g., in the `embedding` store module). This class would handle the core MMR logic, taking a list of `EmbeddingMatch` candidates and returning a re-ranked list.
2.  **Modify `ReRankingContentAggregator`**: The existing `reRankAndFilter` method in `ReRankingContentAggregator` would be updated to perform the MMR re-ranking.
3.  **MMR Algorithm Logic**:
    * The `scoringModel.scoreAll()` method would be used to calculate the **relevance score** (`Sim(D_i, Q)`) for all candidate documents.
    * The embeddings already present in the `Content` objects (retrieved from the embedding store) would be used to calculate the **diversity score** (`max(Sim(D_i, D_j))`) between candidate documents and the documents already selected in the re-ranked list.
    * The final MMR score would be calculated using the formula: `MMR = λ * Relevance - (1 - λ) * Diversity`.
4.  **Configuration**: New optional parameters (`lambda`, `maxResults`) should be added to the `ReRankingContentAggregator` constructor and `builder` to allow users to control the trade-off between relevance and diversity. A default value for `lambda` (e.g., 0.5 or 0.7) can be provided for ease of use.

### Describe alternatives you've considered

One alternative would be to leave the `ReRankingContentAggregator` as-is and instead create a separate `MmrContentAggregator`. However, this would lead to code duplication (e.g., for RRF fusion logic) and a less cohesive architecture.

Another alternative is to implement a simple diversification strategy, such as grouping similar documents and picking one representative from each group. However, MMR is a well-established and robust algorithm that offers a more granular and dynamic way to balance relevance and diversity. Integrating MMR directly into the re-ranking step of `ReRankingContentAggregator` is the most logical and powerful approach.

### Additional context

I previously submitted a Pull Request related to this feature to the core library, but I voluntarily closed it to re-evaluate the approach and align with the project's contribution process. I have since reviewed the guidelines and am now starting this process in the `langchain4j-community` repository.

I have already developed a working implementation of a low-level `MmrReranker` utility class and corresponding unit tests that validate its behavior under various `lambda` values. While this utility currently uses the initial retrieval score for relevance, the final integration within `ReRankingContentAggregator` will be designed to use the more sophisticated score from the `ScoringModel` as the basis for the MMR relevance score. This approach ensures both a robust utility and a powerful final integration. I am ready to open a new pull request in the community repository to get started. I am open to feedback on the best way to integrate this feature and its API design.

can019:
## Issue
Closes #325

## Change
Integrates Maximum Marginal Relevance (MMR) algorithm into `ReRankingContentAggregator` to balance relevance and diversity when selecting documents for LLM context.

### Key Components Added:
1. **`MmrSelector`** - Low-level utility class implementing MMR algorithm
   - Stateless utility operating on `EmbeddingMatch` objects
   - Configurable lambda parameter (0.0-1.0) for relevance vs diversity balance
   - Uses existing scores when available, falls back to cosine similarity calculation
   - Comprehensive parameter validation and edge case handling

2. **Enhanced `ReRankingContentAggregator`**
   - New optional constructor parameter `lambda` for MMR configuration
   - Builder pattern support with `.lambda(double)` method
   - MMR re-ranking integrated into existing `reRankAndFilter` method
   - Backward compatibility maintained (default behavior unchanged when lambda not specified)

### Algorithm Implementation:
- MMR formula: `MMR = λ × Relevance - (1-λ) × max(Similarity(candidate, selected))`
- Uses `ScoringModel` scores for relevance calculation
- Uses cosine similarity between embeddings for diversity measurement
- Iterative selection process for optimal relevance-diversity balance

### Benefits:
- Reduces redundancy in selected documents
- Improves diversity while maintaining relevance
- Provides better LLM context through balanced document selection
- Maintains full backward compatibility

## General checklist
- [ ] I have manually run all the unit tests in *all* modules, and they are all green

### DIFF
--- a/langchain4j-community-core/src/main/java/dev/langchain4j/community/store/embedding/MmrSelector.java
+++ b/langchain4j-community-core/src/main/java/dev/langchain4j/community/store/embedding/MmrSelector.java
@@ -1,9 +1,8 @@
 package dev.langchain4j.community.store.embedding;
 
+import static dev.langchain4j.internal.Utils.isNullOrEmpty;
 import static dev.langchain4j.store.embedding.CosineSimilarity.between;
 import static java.util.Comparator.comparingDouble;
-import static java.util.Objects.isNull;
-import static java.util.Objects.nonNull;
 
 import dev.langchain4j.data.embedding.Embedding;
 import dev.langchain4j.store.embedding.EmbeddingMatch;
@@ -51,7 +50,7 @@ public final class MmrSelector {
         validateParameters(queryEmbedding, lambda, maxResults);
 
         // Handle edge cases
-        if (isNull(candidates) || candidates.isEmpty() || maxResults <= 0) {
+        if (isNullOrEmpty(candidates) || maxResults <= 0) {
             return new ArrayList<>();
         }
 
@@ -67,7 +66,7 @@ public final class MmrSelector {
      * Validates the input parameters for the MMR selection.
      */
     private static void validateParameters(Embedding queryEmbedding, double lambda, int maxResults) {
-        if (isNull(queryEmbedding)) {
+        if (queryEmbedding == null) {
             throw new IllegalArgumentException("Query embedding cannot be null");
         }
         if (lambda < MIN_LAMBDA || lambda > MAX_LAMBDA) {
@@ -96,7 +95,7 @@ public final class MmrSelector {
         while (selected.size() < maxResults && !remaining.isEmpty()) {
             EmbeddingMatch<T> bestCandidate = findBestMmrCandidate(queryEmbedding, remaining, selected, lambda);
 
-            if (nonNull(bestCandidate)) {
+            if (bestCandidate != null) {
                 selected.add(bestCandidate);
                 remaining.remove(bestCandidate);
             } else {
--- a/langchain4j-community-core/src/main/java/dev/langchain4j/community/store/embedding/MmrSelector.java
+++ b/langchain4j-community-core/src/main/java/dev/langchain4j/community/store/embedding/MmrSelector.java
@@ -1,0 +1,185 @@
+package dev.langchain4j.community.store.embedding;
+
+import static dev.langchain4j.store.embedding.CosineSimilarity.between;
+import static java.util.Comparator.comparingDouble;
+import static java.util.Objects.isNull;
+import static java.util.Objects.nonNull;
+
+import dev.langchain4j.data.embedding.Embedding;
+import dev.langchain4j.store.embedding.EmbeddingMatch;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.OptionalDouble;
+
+/**
+ * A utility class that implements the Maximum Marginal Relevance (MMR) algorithm
+ * for selecting diverse and relevant results from a list of embedding matches.
+ * <p>
+ * MMR balances relevance to a query with diversity among the selected results using the formula:
+ * MMR = λ × Relevance(candidate, query) - (1-λ) × max(Similarity(candidate, selected))
+ * <p>
+ * This class is designed to be a low-level, stateless utility that operates on embeddings,
+ * making it reusable across different modules and applications.
+ */
+public final class MmrSelector {
+
+    private static final double INITIAL_MMR_SCORE = -1.0;
+    private static final double INITIAL_DIVERSITY_SCORE = 0.0;
+    private static final double MIN_LAMBDA = 0.0;
+    private static final double MAX_LAMBDA = 1.0;
+
+    private MmrSelector() {}
+
+    /**
+     * Selects a subset of embedding matches using the MMR algorithm to balance relevance and diversity.
+     *
+     * @param queryEmbedding  The embedding of the query for relevance calculation.
+     * @param candidates      The list of candidate matches to select from, typically sorted by relevance.
+     * @param maxResults      The maximum number of results to return.
+     * @param lambda          A value between 0 and 1 (inclusive) that balances relevance and diversity.
+     *                        Higher lambda (e.g., 0.7-0.8) prioritizes relevance,
+     *                        lower lambda (e.g., 0.3-0.4) prioritizes diversity.
+     *                        A value of 1.0 is equivalent to standard relevance-based ranking.
+     * @param <T>             The type of the content that has been embedded.
+     * @return A new list of selected embedding matches ordered by MMR score.
+     * @throws IllegalArgumentException if lambda is not between 0.0 and 1.0 (inclusive),
+     *                                  if queryEmbedding is null, or if maxResults is negative.
+     */
+    public static <T> List<EmbeddingMatch<T>> select(
+            Embedding queryEmbedding, List<EmbeddingMatch<T>> candidates, int maxResults, double lambda) {
+
+        validateParameters(queryEmbedding, lambda, maxResults);
+
+        // Handle edge cases
+        if (isNull(candidates) || candidates.isEmpty() || maxResults <= 0) {
+            return new ArrayList<>();
+        }
+
+        // If we need more results than available, return all candidates
+        if (candidates.size() <= maxResults) {
+            return new ArrayList<>(candidates);
+        }
+
+        return performMmrSelection(queryEmbedding, candidates, maxResults, lambda);
+    }
+
+    /**
+     * Validates the input parameters for the MMR selection.
+     */
+    private static void validateParameters(Embedding queryEmbedding, double lambda, int maxResults) {
+        if (isNull(queryEmbedding)) {
+            throw new IllegalArgumentException("Query embedding cannot be null");
+        }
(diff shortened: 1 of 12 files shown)


langchain4j-community-core/src/main/java/dev/langchain4j/community/store/embedding/MmrSelector.java
```diff
@@ -0,0 +1,185 @@
...
+     * @throws IllegalArgumentException if lambda is not between 0.0 and 1.0 (inclusive),
+     *                                  if queryEmbedding is null, or if maxResults is negative.
+     */
+    public static <T> List<EmbeddingMatch<T>> select(
+            Embedding queryEmbedding, List<EmbeddingMatch<T>> candidates, int maxResults, double lambda) {
+
+        validateParameters(queryEmbedding, lambda, maxResults);
+
+        // Handle edge cases
+        if (isNull(candidates) || candidates.isEmpty() || maxResults <= 0) {
```
Martin7-1:
```suggestion
        if (isNullOrEmpty(candidates) || maxResults <= 0) {
```

langchain4j-community-core/src/main/java/dev/langchain4j/community/store/embedding/MmrSelector.java
```diff
@@ -0,0 +1,185 @@
...
+
+        // Pre-sort candidates by relevance score (descending) for better initial ordering
+        remaining.sort(
+                comparingDouble((EmbeddingMatch<T> match) -> match.score()).reversed());
+
+        // Iteratively select the best candidate based on MMR score
+        while (selected.size() < maxResults && !remaining.isEmpty()) {
+            EmbeddingMatch<T> bestCandidate = findBestMmrCandidate(queryEmbedding, remaining, selected, lambda);
+
+            if (nonNull(bestCandidate)) {
```
Martin7-1: I'd like not to use `nonNull` or `isNull` in this way, as they are mainly to used in `java.util.function.Predicate`

can019: I agree with this

can019:
In the process of extending the existing `RerankAggregator` to apply MMR, we went through the following considerations and decisions:

---

**1. Implementation Approach**  
- The existing `RerankAggregator` resides in the `core` module. Candidate extension approaches included:  
  1) Copy and modify  
  2) Inheritance  
  3) Use the Decorator pattern to wrap `ReRankAggregator`  
  4) Directly implement `ContentAggregator`  
- Both inheritance and Decorator have drawbacks such as difficulty using `ScoreModel` and complicated Builder structure  
- Due to integration test environment issues, **option 4 (direct implementation)** was ultimately chosen

---

**2. Handling Embedding Information**  
- Existing Retrievers do not return embedding information  
- To apply MMR, embeddings are stored in `TextSegment` metadata  
- Since the core class’s static initializer limits allowed metadata types, a workaround of **Base64 encoding → String storage** was used  
- The reason is performance optimization by reusing embeddings already returned by the Retriever  
- However, considering the meaning of Aggregator and potential embedding model mismatches, **re-embedding at the MMR stage** is more reliable, though it incurs embedding cost

---

**3. Embedding Model Consistency**  
- For performance optimization, the embedding model used for user query embeddings and vector storage data embeddings should be identical  
- Optimistically, even if multiple Retrievers exist, they might share the same vector storage and embedding model  
- Pessimistically, if there are N Retrievers, the user query might need to be embedded N times by different models, each followed by vector search

---

**4. Scoring Model Application and MMRSelector Abstraction Needs**  
- Theoretically, MMR only requires **query-document similarity scores** and **document-document similarity scores** to operate  
- However, the current `MmrSelector` implementation is effectively **tied (sticky) to embedding-based cosine similarity calculations**  
- To apply a Scoring Model or alternative similarity calculations, this dependency must be removed, and the architecture improved by raising the abstraction level to allow injection of scoring logic  
- Practically, using a Cross-Encoder for query-document similarity and cosine similarity for document-document similarity is appropriate  
- This implementation was developed based on embedding similarity for validation purposes

---

**5. Others**  
- Separation of Configure / Logging was considered but not done due to the large amount of code written and the need to confirm direction (also, many files were already separated due to strategy implementation)

---

**Request**  
While this implementation prioritizes validation and performance optimization,  
- Re-embedding based MMR  
- Abstraction improvements to integrate Scoring Models  
- Consistency guarantees between Retriever and Embedding models  
are directions we would like to get feedback on.

Martin7-1:
Thanks for your contribution! I'm too busy this week. Will check your comment this weekend :)

can019:
> Thanks for your contribution! I'm too busy this week. Will check your comment this weekend :)

Of course! Please take your time :)

Martin7-1:
> Directly implement ContentAggregator

Yeah, I think directly implement is the best way.

> Since the core class’s static initializer limits allowed metadata types, a workaround of Base64 encoding → String storage was used.

Hmm...This is indeed a problem, one workaround is to provide another `Content` implementation in `langchain4j-community-core`, like `EmbeddingContent`. WDYT?

> * Re-embedding based MMR
> * Abstraction improvements to integrate Scoring Models
> * Consistency guarantees between Retriever and Embedding models

* For `ScoringModel` integration, I think the idea is good , as for now most retriever uses hybrid(token-level and semantic-level) retriever. So just go ahead to improve `MmrSelector`.
* For consistency guarantees between `Retriever` and `EmbeddingModel`, I'm not sure I understand it fully. For now, users can decide which `EmbeddingStore` and which `EmbeddingModel` to use, so what is the question about **consistency**? I think it depends on users decision.

> Separation of Configure / Logging was considered but not done due to the large amount of code written and the need to confirm direction (also, many files were already separated due to strategy implementation)

Yeah. If you can split it into multiple PRs (like Neo4j with a roadmap to continuously optimize), that would greatly reduce my workload, which I would greatly appreciate.

Martin7-1:
BTW, does this PR can merged now? I think everything is good :)

can019:
> Hmm...This is indeed a problem, one workaround is to provide another Content implementation in langchain4j-community-core, like EmbeddingContent. WDYT?

Implementing `EmbeddingContent` sounds good. In that case, should we just provide a separate `EmbeddingStoreContentRetriever` in the community module as we do now?


> For `ScoringModel` integration, I think the idea is good , as for now most retriever uses hybrid(token-level and semantic-level) retriever. So just go ahead to improve `MmrSelector`.

Thanks, got it.

> For consistency guarantees between `Retriever` and `EmbeddingModel`, I'm not sure I understand it fully. For now, users can decide which `EmbeddingStore` and which `EmbeddingModel` to use, so what is the question about consistency? I think it depends on users decision.

For example, if we have **Vector Storage1: Embedding ModelA** and **Vector Storage2: Embedding ModelB**, diversity calculation across the two could be problematic.

Since there seem to be quite a few such cases, I was considering simply outputting a warning and then proceeding with the retriever results and embedding model provided by the user as-is.

can019:
> BTW, does this PR can merged now? I think everything is good :)

Although integration tests are necessary, the cases I had in mind are covered for now!

Martin7-1:
> Since there seem to be quite a few such cases, I was considering simply outputting a warning and then proceeding with the retriever results and embedding model provided by the user as-is.

Yeah, I think it's ok.

Martin7-1 (approved):
@can019 Thank you! Great Job!

PR #357, opened 2025-08-27
Update OCI GenAI tests as Token Usage issue in OCI SDK is fixed

danielkec:
## Change
* Don't set empty maps to SDK builders
* OCI GenAI Generic models token usage fix
* Tests update
* onPartialToolCall support

## General checklist


models/langchain4j-community-oci-genai/src/main/java/dev/langchain4j/community/model/oracle/oci/genai/OciGenAiChatModel.java
```diff
@@ -42,6 +43,7 @@
 public class OciGenAiChatModel extends BaseGenericChatModel<OciGenAiChatModel> implements ChatModel {
 
     private static final Logger LOGGER = LoggerFactory.getLogger(OciGenAiChatModel.class);
+    private static final TokenUsage UNKNOWN_TOKEN_USAGE = new TokenUsage(1, 1, 2);
```
Martin7-1: I think we don't need to set `inputToken`, `outputToken` and `totalToken` as they will confuse the actual token usage, just use: `TokenUsage UNKNOWN_TOKEN_USAGE = new TokenUsage()` is enough I think.

Martin7-1 (commented):
@danielkec Thank you!

models/langchain4j-community-oci-genai/src/main/java/dev/langchain4j/community/model/oracle/oci/genai/OciGenAiChatModel.java
danielkec: I wanted to do that but hitting NPE here in the AbstractBaseChatModelIT based tests:
https://github.com/langchain4j/langchain4j/blob/main/langchain4j-core/src/test/java/dev/langchain4j/model/chat/common/AbstractBaseChatModelIT.java#L1754

Martin7-1: Hmm... What about overriding `assertTokenUsage()` and make it return `false` for now?

BTW, I'm wondering why oci will return a "null" token usage, from your PR description it looks like OCI SDK has fixed this issue.

danielkec: Its still missing for streaming generic models, you are right skipping token usage asserting in that case

PR #445, opened 2025-11-06
Add support for creating SQL Server embedding stores without requiring the embeddings table to be recreated.

odelpozo-denodo:
## Change
With this enhancement, the user of the `langchain4j-community-sqlserver` module will be able to create a `SQLServerEmbeddingStore` **regardless of the existence of the embeddings table**.

If the specified table does not exist, it will be automatically created. Otherwise, the embedding store will utilise the existing table.

Added the option `dev.langchain4j.store.embedding.sqlserver.CreateOption#CREATE_IF_NOT_EXISTS`.

## General checklist


Martin7-1 (approved):
@odelpozo-denodo Thank you!

Issue #379, opened 2025-09-23
[FEATURE] Use UnifiedJedis instead of JedisPooled interface in RedisEmbeddingStore


smilyanovr:
## Is your feature request related to a problem? Please describe.
UnifiedJedis is a common interface which allows interaction with all kind of Redis installations - Sentinel, Cluster, single instance.

## Describe the solution you'd like
Update the type of `RedisEmbeddingStore.client` field to UnifiedJedis.

Martin7-1:
@smilyanovr Hi! Thanks for your feature request, and welcome to contribute :)

XL-Zhao-23:
@smilyanovr  Are you working on this issue?

radosmilyanov-kubit:
@XL-Zhao-23 sorry, I haven't started yet. Last weeks are quite busy... I cannot commit on any timeframe at the moment. I guess I will find some free time in December.

riferrei:
I was able to work on this [here](https://github.com/langchain4j/langchain4j-community/pull/449). I'd love to hear thoughts about using this PR. I think the idea of using UnifiedJedis is simply great, as it opens up new possibilities for connectivity, aligning with where Redis is headed.

PR #449, opened 2025-11-07
Migrate from JedisPooled to UnifiedJedis and Upgrade Jedis to 7.0.0

riferrei:
## Issue

Closes https://github.com/langchain4j/langchain4j-community/issues/379

## Change
This PR updates the client connectivity approach for both the store and memory implementations of Redis. It utilizes the new UnifiedJedis interface, which is the official method for handling Redis connections, whether they are sentinel-based, clustered, or single-node.

## General checklist

- [x] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

### DIFF
--- a/embedding-stores/langchain4j-community-redis/src/main/java/dev/langchain4j/community/store/embedding/redis/RedisEmbeddingStore.java
+++ b/embedding-stores/langchain4j-community-redis/src/main/java/dev/langchain4j/community/store/embedding/redis/RedisEmbeddingStore.java
@@ -79,7 +79,7 @@ public class RedisEmbeddingStore implements EmbeddingStore<TextSegment> {
     /**
      * Creates an instance of RedisEmbeddingStore
      *
-     * @param client         Jedis client
+     * @param client         Instance of a UnifiedJedis client
      * @param indexName      The name of the index (optional). Default value: "embedding-index".
      * @param prefix         The prefix of the key which should end with a colon (e.g., "embedding:") (optional). Default value: "embedding:".
      * @param dimension      Embedding vector dimension


riferrei:
I ended up doing the same with [this other PR](https://github.com/langchain4j/langchain4j-community/pull/449).

Martin7-1 (approved):
@riferrei Thanks for your contribution! All changes look good to me :)

BTW, do you think that we should remove all depreacted parameters in builder and constructor and let users manage `UnifiedJedis` themselves?

riferrei:
> BTW, do you think that we should remove all deprecated parameters in builder and constructor and let users manage `UnifiedJedis` themselves?

That's a great question. On the one hand, the code's implementation mimics the way LangChain-Py was created, allowing developers to interact with the API without needing to delve into the instantiation details of the subsystem. Redis, in this case, is the subsystem. However, I can see value in letting developers handle the UnifiedJedis instance themselves, which would allow them to make decisions like:

- Decide which RESP version to use
- Decide to enable client-side caching
- Fine-tune the client behavior in general

So both case has advantages. The best route is to allow both operations. Perhaps create an overloaded constructor that only accepts the UnifiedJedis instance, and leave the existing one without the UnifiedJedis parameter. As for the builder pattern, same thing. Leave it, but without the UnifiedJedis option.

What do you think?

Martin7-1:
I think it's great!

> As for the builder pattern, same thing. Leave it, but without the UnifiedJedis option.

I'm a little confused about this. Why would you remove the `UnifiedJedis` option from the builder pattern? Perhaps just keep it in the original builder or use two extend builders to distiguish them is a good way:

```
public static class Builder {
    // Some common attributes
}
```

For UnifiedJedis constructor:

```
public static class UnifiedJedisBuilder extend Builder {
    
    private UnifiedJedis unifiedJedis;

    public UnifiedJedisBuilder unifiedJedis(UnifiedJedis unifiedJedis) {
        this.unifiedJedis = unifiedJedis;
        return this;
    }

    public RedisEmbeddingStore build() {
        return ...
    }
}
```

For `host` and `port` constructor:

```
public static class UnifiedJedisBuilder extend Builder {
    
    private String host;
    private String port;

    // builder method

    public RedisEmbeddingStore build() {
        return ...
    }
}
```

WDYT? Would you like to help with this in this PR?

riferrei:
I agree with your comments. I made some changes to the code and added a streamlined way to handle all edge cases, as well as removed redundancy in the index/filterMapper creation from the constructor.

Martin7-1 (approved):
@riferrei Great job! Thank you!

PR #459, opened 2025-11-13
Add support for jvector as an embedding store

odysseaspenta:
## Issue

Closes #1965 from the langchain4j project

## Change

  This module provides a LangChain4j EmbeddingStore implementation backed by https://github.com/jbellis/jvector, which is a pure Java embedded vector search engine.

  Key Features
  - High-performance vector search: Uses graph-based indexing for fast approximate nearest neighbor (ANN) search
  - Flexible storage: In-memory by default with optional persistent storage to disk. Note that this store does not support storing metadata along with the embeddings since jvector doesn't support the storage of metadata.
  - Thread-safe: Concurrent read/write operations with automatic index management
  - Configurable: Supports multiple similarity functions (DOT_PRODUCT, COSINE, EUCLIDEAN) and tunable performance parameters (maxDegree,  beamWidth, etc.)
  - Dynamic updates: Add and remove embeddings at runtime with automatic index rebuilding. Note that this embedding store runs in memory and does not utilize a remote server

## General checklist

- [X] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [X] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

- [X] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data persisted using the latest released version of LangChain4j

### DIFF
--- a/embedding-stores/langchain4j-community-jvector/src/main/java/dev/langchain4j/community/store/embedding/jvector/JVectorEmbeddingStore.java
+++ b/embedding-stores/langchain4j-community-jvector/src/main/java/dev/langchain4j/community/store/embedding/jvector/JVectorEmbeddingStore.java
@@ -32,7 +32,6 @@ import java.io.FileOutputStream;
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
-import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
@@ -480,7 +479,7 @@ public class JVectorEmbeddingStore implements EmbeddingStore<TextSegment> {
         }
 
         // Save the graph index
-        Path graphPath = Paths.get(persistencePath + ".graph");
+        Path graphPath = Path.of(persistencePath + ".graph");
         Files.createDirectories(graphPath.getParent());
 
         RandomAccessVectorValues vectorValues = new ListRandomAccessVectorValues(vectors, dimension);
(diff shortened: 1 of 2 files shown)


Martin7-1:
@odysseaspenta Thanks for your contribution!

I'm not very familiar with JVector, and one of my doubts is, if it need to rebuild the index before search, will this cause a significant decrease in search performance?

Or, should we provide a parameter to control for eagle-rebuild(rebuild during add and remove) or lazy-rebuild(rebuild during search)?

As for `indexLock`, I think its introduction will lead to uncertainty in `JVectorEmbeddingStore` under different scenarios. e.g. Will a large number of concurrent searches cause add/remove operations to fail to obtain `indexLock` and thus block? I'm unsure if this is a responsibility that `langchain4j-community` needs to address.

odysseaspenta:
I was mainly focused on the scenario where a large number of vectors are added to the store up front and then the index is used for search operations, so that it is only the first search request that will incur slow performance. I like the idea of adding a parameter to control eager vs lazy rebuilds of the index.

JVector uses  integers as ids for vectors (ordinals) and doesn't store the text associated with the embedding vector so the implementation uses maps to maintain associations between langchain4j's expectations and what jvector supports. I used the locking to keep the vector data managed by jvector in sync with the maps maintained by the embedding store implementation. How would you suggest we eliminate the locking?

Martin7-1:
> How would you suggest we eliminate the locking?

As for now, I think we can do nothing with the locking, until `JVector` improve and can store custom id and text segment.

Issue #433, opened 2025-11-02
[FEATURE] RedisChatMemoryStore supports native set and get commands


Xy-hub:
## Is your feature request related to a problem? Please describe.
No

## Describe the solution you'd like
Users can switch the implementation method of reading and writing data

## Describe alternatives you've considered
I hope the implementation method in RedisChatMemoryStore can be made configurable, because based on my personal experience, the low-version native Redis does not integrate the RedisJson plugin, and I need to install this plugin separately, which adds some difficulty to my use of RedisChatMemoryStore. Therefore, I hope to add a new configuration option that uses the RedisJson commands by default, but also allows setting up to use the native String set and get commands for read and write operations. 
Like this:
langchain4j:
  open-ai:
    redis-chat-memory:
      method: json or string

Thank you

## Additional context
No

Martin7-1:
@Xy-hub  Good point I think! Would you like to contribute to it?

Xy-hub:
@Martin7-1 OK, I will try it.

PR #461, opened 2025-11-16
RedisChatMemoryStore is compatible with both RedisJson format and String format

Xy-hub:
- RedisChatMemoryStore Support JSON Data Type and String Data Type
- Default use JSON Data Type

## Issue

Feature #433 

## General checklist

### DIFF
--- a/content-retrievers/langchain4j-community-neo4j-retriever/pom.xml
+++ b/content-retrievers/langchain4j-community-neo4j-retriever/pom.xml
@@ -14,7 +14,7 @@
 
     <properties>
         <neo4j.version>6.0.2</neo4j.version>
-        <neo4j.cypher.dsl.version>2025.1.0</neo4j.cypher.dsl.version>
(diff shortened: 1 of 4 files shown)


Martin7-1 (commented):
@Xy-hub Thanks for your contribution! For the code maintainability, could you please refactor the code and do not use `if-else` pattern here.

My concern is that if new storage types are introduced in the future, there will be even more if-else statements here.

What about defining a new interface named `RedisOperatons` and have `add`, `remove` and so on, and then implement it with `JSONRedisOperations` and `StringRedisOperations`.

After that, I think we can integrate it into `RedisChatMemoryStore` by a simple member variable. And in the future, we can just implement a new `RedisOperations` type and do not need change the code of `RedisChatMemoryStore`.

WDYT?

Xy-hub:
@Martin7-1 Thank you very much for your suggestions. I will proceed to refactor this part of the code accordingly.
Thanks

Xy-hub:
@Martin7-1 I am very happy to contribute my own strength, and I hope the community will become better and better!
Thanks

PR #536, opened 2026-01-08
feat(s3-vectors): Add Amazon S3 Vectors embedding store

ss-bhatt:
## Issue

Closes # https://github.com/langchain4j/langchain4j/issues/4192

## Change

Adds Amazon S3 Vectors embedding store integration.

**What's included:**
- `S3VectorsEmbeddingStore` with cosine/euclidean distance support
- `S3VectorsMetadataFilterMapper` for MongoDB-style filter syntax
- Full test coverage (26 unit tests + 2 IT classes)

**Notes:**
- Extends `EmbeddingStoreIT` instead of `EmbeddingStoreWithFilteringIT` since S3 Vectors uses ANN search with ~6% score variance. The FilteringIT base class hardcodes 0.01% tolerance which doesn't account for ANN approximation. Same pattern as `Cassandra`.
- AWS SDK version pinned to 2.41.0 in module POM as the `s3vectors` artifact doesn't exist in older versions
- `removeAll(Filter)` throws `UnsupportedOperationException`. The S3 Vectors API only supports deletion by key, not by filter
- Default text metadata key is `_page_content` to match Python langchain-aws for cross language interops.

**Testing:**
- Integration tests require AWS credentials (same as `Bedrock` module). Verified against real S3 Vectors. All ITs pass.

**Additional Context:**
- https://aws.amazon.com/s3/features/vectors/
- https://aws.amazon.com/blogs/aws/introducing-amazon-s3-vectors-first-cloud-storage-with-native-vector-support-at-scale/
- https://aws.amazon.com/about-aws/whats-new/2025/07/amazon-s3-vectors-preview-native-support-storing-querying-vectors/

## General checklist

- [X] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [X] I have added a `{NameOfIntegration}EmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

### DIFF
--- a/embedding-stores/langchain4j-community-s3-vectors/src/main/java/dev/langchain4j/community/store/embedding/s3/S3VectorsEmbeddingStore.java
+++ b/embedding-stores/langchain4j-community-s3-vectors/src/main/java/dev/langchain4j/community/store/embedding/s3/S3VectorsEmbeddingStore.java
@@ -484,12 +484,12 @@ public class S3VectorsEmbeddingStore implements EmbeddingStore<TextSegment>, Aut
     }
 
     private Document toDocument(Object value) {
-        if (value instanceof String) {
-            return Document.fromString((String) value);
-        } else if (value instanceof Number) {
-            return Document.fromNumber(((Number) value).toString());
-        } else if (value instanceof Boolean) {
-            return Document.fromBoolean((Boolean) value);
+        if (value instanceof String string) {
+            return Document.fromString(string);
+        } else if (value instanceof Number number) {
+            return Document.fromNumber(number.toString());
+        } else if (value instanceof Boolean b) {
+            return Document.fromBoolean(b);
         } else {
             return Document.fromString(String.valueOf(value));
         }
@@ -562,17 +562,5 @@ public class S3VectorsEmbeddingStore implements EmbeddingStore<TextSegment>, Aut
         if (s3VectorsClient != null) {
             s3VectorsClient.close();
         }
-    }
-
-    S3VectorsClient getS3VectorsClient() {
-        return s3VectorsClient;
-    }
-
-    String getVectorBucketName() {
-        return vectorBucketName;
-    }
-
-    String getIndexName() {
-        return indexName;
     }
 }
--- a/embedding-stores/langchain4j-community-s3-vectors/src/main/java/dev/langchain4j/community/store/embedding/s3/S3VectorsMetadataFilterMapper.java
+++ b/embedding-stores/langchain4j-community-s3-vectors/src/main/java/dev/langchain4j/community/store/embedding/s3/S3VectorsMetadataFilterMapper.java
@@ -33,28 +33,28 @@ class S3VectorsMetadataFilterMapper {
     static Document map(Filter filter) {
         if (filter == null) {
             return null;
-        } else if (filter instanceof IsEqualTo) {
-            return mapEqual((IsEqualTo) filter);
-        } else if (filter instanceof IsNotEqualTo) {
-            return mapNotEqual((IsNotEqualTo) filter);
-        } else if (filter instanceof IsGreaterThan) {
-            return mapGreaterThan((IsGreaterThan) filter);
-        } else if (filter instanceof IsGreaterThanOrEqualTo) {
-            return mapGreaterThanOrEqual((IsGreaterThanOrEqualTo) filter);
-        } else if (filter instanceof IsLessThan) {
-            return mapLessThan((IsLessThan) filter);
-        } else if (filter instanceof IsLessThanOrEqualTo) {
-            return mapLessThanOrEqual((IsLessThanOrEqualTo) filter);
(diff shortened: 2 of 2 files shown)


embedding-stores/langchain4j-community-s3-vectors/src/main/java/dev/langchain4j/community/store/embedding/s3/S3VectorsEmbeddingStore.java
```diff
@@ -0,0 +1,578 @@
...
+                if (value != null) {
+                    metadataMap.put(key, toDocument(value));
+                }
+            });
+        }
+
+        return Document.fromMap(metadataMap);
+    }
+
+    private Document toDocument(Object value) {
```
Martin7-1, line 486: nit:

```java
    private Document toDocument(Object value) {
        if (value instanceof String string) {
            return Document.fromString(string);
        } else if (value instanceof Number number) {
            return Document.fromNumber(number.toString());
        } else if (value instanceof Boolean b) {
            return Document.fromBoolean(b);
        } else {
            return Document.fromString(String.valueOf(value));
        }
    }
```

embedding-stores/langchain4j-community-s3-vectors/src/main/java/dev/langchain4j/community/store/embedding/s3/S3VectorsEmbeddingStore.java
```diff
@@ -0,0 +1,578 @@
...
+    }
+
+    @Override
+    public void close() {
+        if (s3VectorsClient != null) {
+            s3VectorsClient.close();
+        }
+    }
+
+    S3VectorsClient getS3VectorsClient() {
```
Martin7-1: nit: Looks like this method can be removed.

embedding-stores/langchain4j-community-s3-vectors/src/main/java/dev/langchain4j/community/store/embedding/s3/S3VectorsEmbeddingStore.java
```diff
@@ -0,0 +1,578 @@
...
+        return s3VectorsClient;
+    }
+
+    String getVectorBucketName() {
```
Martin7-1: nit: Looks like this method can be removed.

embedding-stores/langchain4j-community-s3-vectors/src/main/java/dev/langchain4j/community/store/embedding/s3/S3VectorsEmbeddingStore.java
```diff
@@ -0,0 +1,578 @@
...
+        return vectorBucketName;
+    }
+
+    String getIndexName() {
```
Martin7-1: nit: Looks like this method can be removed.

embedding-stores/langchain4j-community-s3-vectors/src/main/java/dev/langchain4j/community/store/embedding/s3/S3VectorsMetadataFilterMapper.java
```diff
@@ -0,0 +1,195 @@
...
+ * Maps langchain4j Filter to AWS S3 Vectors filter format.
+ * S3 Vectors uses MongoDB-style filter syntax.
+ */
+class S3VectorsMetadataFilterMapper {
+
+    private S3VectorsMetadataFilterMapper() {
+        // utility class
+    }
+
+    static Document map(Filter filter) {
```
Martin7-1, line 33: nit: it would be better to use pattern match for instanceof

embedding-stores/langchain4j-community-s3-vectors/src/main/java/dev/langchain4j/community/store/embedding/s3/S3VectorsMetadataFilterMapper.java
```diff
@@ -0,0 +1,195 @@
...
+    }
+
+    private static Document mapOr(Or or) {
+        return createLogicalDocument("$or", map(or.left()), map(or.right()));
+    }
+
+    /**
+     * S3 Vectors does not support "$not" operation directly, so we convert to inverse operations.
+     */
+    private static Document mapNot(Not not) {
```
Martin7-1, line 107: nit: it would be better to use pattern match for instanceof

ss-bhatt:
@Martin7-1 : Thank you for the review. Good call on pattern matching. I have applied those changes and removed the unused helper methods. Ready for another look!

Martin7-1:
@ss-bhatt Hi! Could you please add document in langchain4j main repo?

ss-bhatt:
@Martin7-1 : On It. Thanks for the merge.

PR #583, opened 2026-02-26
fix(router): ChatModelWrapper NPE bug

vesense:
## Issue

Closes #

Fix the issue `supportedCapabilities` always throw NPE.

## General checklist

### DIFF
--- a/models/langchain4j-community-model-router/src/main/java/dev/langchain4j/model/router/ChatModelWrapper.java
+++ b/models/langchain4j-community-model-router/src/main/java/dev/langchain4j/model/router/ChatModelWrapper.java
@@ -13,7 +13,6 @@ import dev.langchain4j.model.chat.response.StreamingChatResponseHandler;
 import java.io.Serializable;
 import java.util.ArrayList;
 import java.util.HashMap;
-import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Objects;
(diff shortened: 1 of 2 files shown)


models/langchain4j-community-model-router/src/main/java/dev/langchain4j/model/router/ChatModelWrapper.java
```diff
@@ -91,12 +90,11 @@ public void doChat(ChatRequest chatRequest, StreamingChatResponseHandler handler
 
     @Override
     public Set<Capability> supportedCapabilities() {
-        Set<Capability> capabilities = new HashSet<>();
         if (model != null) {
-            capabilities.addAll(model.supportedCapabilities());
+            return model.supportedCapabilities();
```
Martin7-1, line 94: I think for capabilities, we should take their union rather than just one of them.

Martin7-1 (commented):
@vesense Thanks for your contribution!

models/langchain4j-community-model-router/src/main/java/dev/langchain4j/model/router/ChatModelWrapper.java
vesense: Hi, I thought this point. Finally I kept the behavior like `listeners()` or `provider()` in `ChatModelWrapper `.

```
    public List<ChatModelListener> listeners() {
        if (model != null) {
            return Stream.concat(model.listeners().stream(), listeners.stream()).toList();
        }
        if (streamingModel != null) {
            return Stream.concat(streamingModel.listeners().stream(), listeners.stream())
                    .toList();
        }
        throw new NullPointerException("both model and streamingModel are null");
    }
```

Martin7-1: OK, I got it.

Martin7-1 (approved):
@vesense Thank you!

PR #713, opened 2026-06-28
Oci genai missing tool

ashishbania:
## Issue
Closes #607

## Change
generate uuid and assign in case function call id is null

## General checklist


Martin7-1 (approved):
@ashishbania Thank you!

PR #745, opened 2026-08-10
feat(xinference): Adding thinking support for `ChatModel` and `StreamingChatModel`

### LINKED ISSUE #304: Support reasoning
[#3380](https://github.com/langchain4j/langchain4j/pull/3380)

- [ ] DashScope
- [ ] XInference
- [ ] Qianfan

diegoberriosr:
## Issue

Closes #304 (partially)

## Change

Added support for enabling/disabling reasoning content for the `ChatModel` and `StreamingChatModel` implementations of the Xinference integration.

- `reasoning_content` request parameter can be enabled/disabled using each classes' respective builder.
-  Implemented`AbstractXinferenceThinkingModelInfrastructure` for the new ITs that cover this feature (it requires a little bit more setup, like enabling automatic parsing for the reasoning content in the actual Xinference server).
-  Extended reasoning content support configuration for the integration's Spring Boot starters as well.

**Note**: I did a small refactor of `XinferenceChatModel` and `XinferenceStreamingChatModel` constructors to make the changes non-breaking since it required altering their signatures, and also marked them as deprecated (to favour the builder pattern usage).

## General checklist

- [x] I have added/updated [Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring) (if applicable)


Martin7-1:
@diegoberriosr Hi! Thanks for the contribution! Please let me know if the draft PR can be reviewd

diegoberriosr:
Hey @Martin7-1! 

Apologies for the late response. Yes, the PR should be ready for review now, thank you!

**P.S**: Just for full transparency, I noticed some tests are failing both in `main` and my feature branch so I didn't check the 'module ITs are all passing' item in the PR's template. 

(I think most of the failing cases are tests that were gradually added in the abstract ITs of the core l4j repo but were never adapted for this module in particular, like `XinferenceToolChatModelIT.should_execute_tool_without_parameters`)


image

diegoberriosr:
btw, I think most of them should be relatively straightforward to fix, so I could open a new PR patching them (if you're okay with it ofc)

Martin7-1 (commented):
Thanks for this PR! A few review comments inline. One additional point on test coverage, comparing with `langchain4j-community-dashscope`:

**Missing common ITs (vs dashscope)**

Dashscope implements the common IT suites from the `langchain4j-core` test jar, while xinference currently only has hand-written ITs:

- `QwenChatModelIT extends AbstractChatModelIT` and `QwenStreamingChatModelIT extends AbstractStreamingChatModelIT` (`dev.langchain4j.model.chat.common`) — these bring ~30 shared tests: user/system/multiple messages, `modelName`/`maxOutputTokens`/`stopSequences` handling, tool execution (single/no-arg/parallel, tool choice), JSON response formats, token usage assertions, etc. Other modules (cohere, qianfan, oci-genai) have them too.
- `QwenAiServicesIT` / `QwenAiServicesWithToolsIT` / `QwenStreamingAiServiceIT` (`dev.langchain4j.service.common`).

It would be good to add `XinferenceChatModelIT extends AbstractChatModelIT` and `XinferenceStreamingChatModelIT extends AbstractStreamingChatModelIT` (in the existing `common` test package). Note this likely needs per-request `ChatRequestParameters` support first (there is an existing `TODO: support custom ChatRequestParameters` in `XinferenceChatModel`) — if that's out of scope here, a follow-up issue would be great.

**Streaming reasoning IT: verify `onPartialThinking` emission**

The new `XinferenceThinkingStreamingChatModelIT` only asserts the final `AiMessage.thinking()`. If the `onPartialThinking` emission block were removed, both tests would still pass (the response builder accumulates reasoning independently). Dashscope has the pattern to copy in `QwenStreamingChatModelReasoningIT`:

```java
TestStreamingChatResponseHandler spyHandler = spy(new TestStreamingChatResponseHandler());
// ... enabled case:
verify(spyHandler, atLeastOnce()).onPartialThinking(any());
assertThat(aiMessage.thinking()).isEqualTo(spyHandler.getThinking());
// ... disabled case:
verify(spyHandler, never()).onPartialThinking(any());
```

**Unit tests for the rewritten logic**

`XinferenceStreamingResponseBuilder` and `InternalXinferenceHelper.aiMessageFrom` were rewritten but have no unit tests. Small container-free unit tests (accumulation, empty stream, thinking on/off) would be very valuable here.

models/langchain4j-community-xinference/src/main/java/dev/langchain4j/community/model/xinference/XinferenceStreamingResponseBuilder.java
```diff
@@ -119,37 +124,33 @@ public void append(CompletionResponse partialResponse) {
 
     public ChatResponse build() {
```
Martin7-1, line 125: Silent contract change: `build()` previously returned `null` when nothing was accumulated; it now always returns a `ChatResponse`. Since core `AiMessage.builder()` performs no validation, an empty stream now yields a response wrapping an all-null `AiMessage` (text/thinking/toolExecutionRequests all null).

Impact on the two callers:
- `XinferenceStreamingChatModel`: an empty stream (aborted generation, content filtering) now silently yields a valid response with `text() == null` instead of the previous null signal.
- `XinferenceStreamingLanguageModel` (~line 119): previously NPE-d on `response.aiMessage()` for empty streams; now `Response.from(null, ...)` throws `IllegalArgumentException("content cannot be null")` inside the SSE `onClosed` callback, which is not caught — so neither `onComplete` nor `onError` fires.

Could this be made deliberate — e.g. documented, or the "nothing accumulated" case handled explicitly (error or sentinel) — plus a unit test pinning the behavior?

models/langchain4j-community-xinference/src/main/java/dev/langchain4j/community/model/xinference/InternalXinferenceHelper.java
```diff
@@ -135,17 +135,21 @@ static Tool toToolChoice(ToolSpecification toolSpecification) {
...
-                    .filter(toolCall -> toolCall.getType() == ToolType.FUNCTION)
-                    .map(InternalXinferenceHelper::toToolExecutionRequest)
-                    .toList();
-            return isNullOrBlank(text)
-                    ? AiMessage.from(toolExecutionRequests)
-                    : AiMessage.from(text, toolExecutionRequests);
-        }
-        return AiMessage.from(text);
+
+        List<ToolExecutionRequest> toolExecutionRequests = isNullOrEmpty(toolCalls)
```
Martin7-1, line 141: Silent behavior change: previously, if `toolCalls` was non-empty but every entry was filtered out by the `ToolType.FUNCTION` filter, `AiMessage.from(emptyList)` threw `IllegalArgumentException` — a loud signal that the server sent tool calls we cannot represent. Now the empty list passes through the unvalidated builder, so the tool calls are silently dropped and the caller proceeds as if it were a plain reply — an agent loop could silently stop calling tools.

Suggestion: fail (or at least log) when `toolCalls` is non-empty but `toolExecutionRequests` ends up empty.

Also a minor idiom nit: `.text(isNotNullOrBlank(text) ? text : null)` — the idiom used elsewhere (e.g. `InternalZhipuAiHelper`, and this PR's own `XinferenceStreamingResponseBuilder.build()`) is the flipped form `isNullOrBlank(x) ? null : x`.

models/langchain4j-community-xinference/src/main/java/dev/langchain4j/community/model/xinference/XinferenceStreamingChatModel.java
```diff
@@ -152,6 +183,11 @@ public void doChat(ChatRequest request, StreamingChatResponseHandler handler) {
                         if (isNotNullOrEmpty(content)) {
                             handler.onPartialResponse(content);
                         }
+
+                        String reasoningContent = delta.getReasoningContent();
+                        if (isNotNullOrEmpty(reasoningContent)) {
```
Martin7-1, line 188: Threshold mismatch: partial thinking is emitted under `isNotNullOrEmpty`, but the final `AiMessage.thinking()` is only set when non-**blank** (`isNullOrBlank` in `XinferenceStreamingResponseBuilder.build()`). Whitespace-only reasoning deltas will fire `onPartialThinking` events yet be absent from the final message. Suggest aligning the two checks (and with the `onPartialResponse`/content handling).

models/langchain4j-community-xinference/src/main/java/dev/langchain4j/community/model/xinference/InternalXinferenceHelper.java
```diff
@@ -135,17 +135,21 @@ static Tool toToolChoice(ToolSpecification toolSpecification) {
...
+                ? null
+                : toolCalls.stream()
+                        .filter(toolCall -> toolCall.getType() == ToolType.FUNCTION)
+                        .map(InternalXinferenceHelper::toToolExecutionRequest)
+                        .toList();
+
+        return AiMessage.builder()
+                .text(isNotNullOrBlank(text) ? text : null)
```
Martin7-1, line 149: Subtle behavior change on a degenerate path: the client `AssistantMessage` defaults null content to `""`, and the old code returned `AiMessage.from(text)` unconditionally — so an empty response previously produced `text() == ""`. The new code maps blank to null, so downstream code doing `text().isEmpty()` would now NPE. Is this intentional? If kept, worth pinning with a unit test.

models/langchain4j-community-xinference/src/test/java/dev/langchain4j/community/model/xinference/XinferenceThinkingStreamingChatModelIT.java
```diff
@@ -0,0 +1,58 @@
+package dev.langchain4j.community.model.xinference;
+
+import static org.assertj.core.api.AssertionsForClassTypes.assertThat;
```
Martin7-1, line 3: All other xinference ITs import `org.assertj.core.api.Assertions.assertThat` — please use that here too for consistency (same in `XinferenceThinkingModelIT`).

models/langchain4j-community-xinference/src/test/java/dev/langchain4j/community/model/xinference/XinferenceUtils.java
```diff
@@ -17,6 +17,8 @@ public class XinferenceUtils {
     //    public static final String XINFERENCE_IMAGE = "xprobe/xinference:latest";
 
     public static final String CHAT_MODEL_NAME = "qwen3";
+    public static final String THINKING_MODEL_NAME = "qwen3-thinking";
+    private static final String THINKING_BASE_MODEL_NAME = "qwen3";
```
Martin7-1, line 21: `THINKING_BASE_MODEL_NAME` duplicates the value of the existing `CHAT_MODEL_NAME` ("qwen3"). Could reuse `CHAT_MODEL_NAME` in the launch command — unless the two are deliberately allowed to drift.

Also, optional polish for the two new IT classes: the two tests in each class differ only in `enableThinking(true/false)`, so a small `private static` factory method would remove ~20 lines of duplicated builder setup.

PR #748, opened 2026-08-12
feat(cohere): Implement TokenCountEstimator for Cohere integration

### LINKED ISSUE #747: [FEATURE] Add `/tokenize` endpoint suport for Cohere integration
## Is your feature request related to a problem? Please describe.

Many integrations support estimating tokens in some sort of format via `TokenCountEstimator`, which is very handy in flows where you need to consider the approximate number of tokens the request will consume before executing it.

`langchain4j-community-cohere` currently does not support this functionality, but this could be implemented using the [official API's `/tokenize` endpoint](https://docs.cohere.com/reference/tokenize), since it allows you to send piece of text and get an array with the tokenized representation of the input as a result. 

The length of that array can then be used  for a `TokenCountEstimator` class adapted to the Cohere integration module. 

## Describe the solution you'd like

- As a user of the module, I'd like to be able to estimate the number of tokens a request might consume with an implementation of `TokenCountEstimator` adapted to the Cohere API.

## Describe alternatives you've considered

None

## Additional context

None

diegoberriosr:
## Issue

Closes #747

## Change

Added an implementation of `TokenCountEstimator` using [Cohere's tokenization API](https://docs.cohere.com/reference/tokenize) (`/tokenize` endpoint for V1):

- `CohereTokenCountEstimator` can be configured with multiple configurable parameters  (`modelName`, `maxRetries`, `timeout`, etc) via builder instantiation.
- It uses the `/tokenize` response object `token` field to get the token count for a given string (`token` corresponds to the tokenized version of the input string, so it's size can be used to calculate the number of tokens).
- Since the API endpoint expects raw text to tokenize (unlike Anthropic, which receives the actual message object), the final implementation ends up being similar to [WatsonXTokenCountEstimator](https://github.com/langchain4j/langchain4j/blob/main/langchain4j-watsonx/src/main/java/dev/langchain4j/model/watsonx/WatsonxTokenCountEstimator.java) where you try to extract all relevant text information for each supported type of `ChatMessage`.
- Only `TextContent` is supported for tokenization, other formats like `PdfFileContent` will cause the token estimator to throw an exception.
- For messages with many text possible blocks (like `AiMessage` with `text` and `thinking`), a single chunk of text is sent in the request to avoid calling the API too many times. 

*Full test suite run result for the Cohere integration module*

image

*new ITs run result*

image

## General checklist

### DIFF
--- a/models/langchain4j-community-cohere/src/main/java/dev/langchain4j/community/model/CohereTokenCountEstimator.java
+++ b/models/langchain4j-community-cohere/src/main/java/dev/langchain4j/community/model/CohereTokenCountEstimator.java
@@ -1,0 +1,230 @@
+package dev.langchain4j.community.model;
+
+import static dev.langchain4j.internal.RetryUtils.withRetryMappingExceptions;
+import static dev.langchain4j.internal.Utils.getOrDefault;
+import static dev.langchain4j.internal.Utils.isNotNullOrEmpty;
+import static dev.langchain4j.internal.ValidationUtils.ensureNotBlank;
+
+import dev.langchain4j.community.model.client.CohereClient;
+import dev.langchain4j.community.model.client.tokenize.CohereTokenizeRequest;
+import dev.langchain4j.community.model.client.tokenize.CohereTokenizeResponse;
+import dev.langchain4j.data.message.AiMessage;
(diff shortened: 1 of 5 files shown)


Martin7-1 (approved):
@diegoberriosr Thank you!

PR #752, opened 2026-08-16
fix: bound failover retry in ModelRouter to avoid unbounded recursion

### LINKED ISSUE #751: [BUG] ModelRouter doChat recurses unboundedly on persistent delegate failure -> StackOverflowError
## Describe the bug

`ModelRouter.doChat(...)` (and the streaming `doChat(...)` overload) recursively retries the request **without any bound** whenever a delegate `ChatModel` throws. When the routed model persistently fails and the routing strategy does not mark it failed (for example `LowestTokenUsageRoutingStrategy`, or any strategy with a single route), this recursion never terminates and the JVM crashes with a `StackOverflowError` instead of surfacing the underlying failure.

See https://github.com/langchain4j/langchain4j-community/blob/main/models/langchain4j-community-model-router/src/main/java/dev/langchain4j/model/router/ModelRouter.java

```java
@Override
public ChatResponse doChat(ChatRequest chatRequest) {
    ChatModelWrapper delegate = resolveDelegate(chatRequest);
    try {
        return delegate.chat(chatRequest);
    } catch (NoMatchingModelFoundException e) {
        throw e;
    } catch (Exception e) {
        return doChat(chatRequest);   // <-- unbounded recursion on failure
    }
}
```

The retry seems intended to support failover (see `FailoverStrategy`), but it is uncapped, so a permanently failing delegate causes infinite recursion.

## To reproduce

Construct a `ModelRouter` with a single model whose `doChat` always throws, and any routing strategy that does not mark the model as failed, e.g. `LowestTokenUsageRoutingStrategy`:

```java
ModelRouter router = ModelRouter.builder()
    .addRoutes(new FlakyChatModel())               // doChat always throws
    .routingStrategy(new LowestTokenUsageRoutingStrategy())
    .build();
router.chat(ChatRequest.userMessage("ping"));     // StackOverflowError (unbounded recursion)
```

## Expected behavior

The underlying delegate failure should surface as a normal, bounded exception once the failover attempts are exhausted (one attempt per configured route), rather than recursing until the stack overflows.

## Environment
- langchain4j-community version: `main` (1.19.0-beta29-SNAPSHOT)
- Java version: 17

seanxuu:
## Issue

Fixes #751

## Change

`ModelRouter.doChat(...)` and its streaming overload recursed into themselves **without a bound** every time a delegate `ChatModel` threw. For a permanently-failing route combined with a routing strategy that does not mark the model as failed (e.g. `LowestTokenUsageRoutingStrategy`, or any strategy with a single route), this recursion never terminates and crashes the JVM with a `StackOverflowError` instead of surfacing the underlying failure.

This change bounds the failover retry to the number of configured routes: each route is given a single attempt before the last exception propagates. Both the synchronous and the streaming `doChat` paths are fixed identically. The intended failover behaviour (skip failed models, try the next healthy route) is preserved.

## General checklist
- [x] There are no breaking changes (API, behaviour) — the public `doChat` signatures are unchanged; on success and with a healthy failover the result is identical, only the unbounded failure path is now bounded.
- [x] I have added unit and/or integration tests for my change (`ModelRouterTest`)
- [x] The tests cover both positive and negative cases
- [x] I have manually run all unit tests in the changed module and they pass

## How to test

```shell
./mvnw -pl models/langchain4j-community-model-router test
```

`ModelRouterTest` includes:
- `propagatesFailureInsteadOfRecursingUnboundedly` — a permanently-failing model now surfaces `IllegalStateException` (previously `StackOverflowError`).
- `failsOverToHealthyModel` — a failing route followed by a healthy one routes to the healthy model within the bound.
- `boundedAttemptsWhenStrategyKeepsSelectingSameFailingModel` — with a single route it terminates after one attempt and propagates the failure.
- `successfulCallDelegatesToRoute` — happy path unchanged.

### DIFF
--- a/models/langchain4j-community-model-router/src/main/java/dev/langchain4j/model/router/ModelRouter.java
+++ b/models/langchain4j-community-model-router/src/main/java/dev/langchain4j/model/router/ModelRouter.java
@@ -65,26 +65,43 @@ public class ModelRouter implements ChatModel, StreamingChatModel {
 
     @Override
     public ChatResponse doChat(ChatRequest chatRequest) {
+        // Bound the failover retry to the number of configured routes: each route is
+        // given a single attempt before the original failure propagates. Without this
+        // bound, a persistently-failing delegate would cause unbounded recursion and a
+        // StackOverflowError instead of surfacing the underlying failure.
+        return doChatInternal(chatRequest, this.routes.size());
+    }
+
+    private ChatResponse doChatInternal(ChatRequest chatRequest, int attemptsLeft) {
         ChatModelWrapper delegate = resolveDelegate(chatRequest);
         try {
-            ChatResponse response = delegate.chat(chatRequest);
-            return response;
+            return delegate.chat(chatRequest);
         } catch (NoMatchingModelFoundException e) {
             throw e;
         } catch (Exception e) {
-            return doChat(chatRequest);
+            if (attemptsLeft <= 1) {
+                throw e;
+            }
+            return doChatInternal(chatRequest, attemptsLeft - 1);
         }
     }
 
     @Override
     public void doChat(ChatRequest chatRequest, StreamingChatResponseHandler handler) {
+        doChatInternal(chatRequest, handler, this.routes.size());
+    }
+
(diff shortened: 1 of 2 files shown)


Martin7-1 (approved):
@seanxuu Thank you! All changes look good to me.

PR #755, opened 2026-08-18
Add Amazon DynamoDB embedding store

riferrei:
## Change

Adds a new embedding store integration: **`langchain4j-community-dynamodb`**, an `EmbeddingStore<TextSegment>` backed by Amazon DynamoDB's **native vector search**. This is a strategic addition as it complements the already present support for Amazon S3 Vectors. While Amazon S3 Vectors is recommended for infrequent queries, Amazon DynamoDB is recommended for scenarios where single-digit millisecond latency at 99%+ recall is needed.

Highlights:

- **Native vector search.** Each embedding is stored as a single DynamoDB item whose partition key is the embedding id; the vector is stored as a list of numbers under a configurable attribute and indexed by a DynamoDB vector index. Supports the `COSINE`, `EUCLIDEAN` and `DOT_PRODUCT` distance functions and vectors up to 4096 dimensions. `maxResults` is capped at DynamoDB's `TopK` limit of 100.
- **Metadata.** Stored per-attribute. Integer/long values use the native DynamoDB `Number` type; float and double values are stored as tagged strings, because DynamoDB's `Number` type underflows on subnormal magnitudes (e.g. `Double.MIN_VALUE`). A single `DynamoDbAttributeCodec` is shared by the write path and the filter mapper so equality filters on float/double values still match the persisted representation.
- **Filtering.** DynamoDB vector search only filters on attributes declared in the index search schema (the `HASH` key and `INLINE_FILTER` attributes, fixed at index creation) and only with equality. Accordingly, `IsEqualTo` filters (optionally combined with `And`) are mapped to a `SearchConditionExpression`; other filter types throw `UnsupportedOperationException`. `removeAll(Filter)` is likewise unsupported.
- **Robustness.** Metadata keys colliding with the reserved id/vector/text attributes are rejected on write; `BatchWriteItem` retries unprocessed items with capped exponential backoff; table activation waits with a timeout; and a caller-supplied `DynamoDbClient` is not closed by the store.

Also registers the module in the root `pom.xml` and `langchain4j-community-bom/pom.xml`.

## General checklist

- [ ] I have added an example in the [[examples repo](https://github.com/langchain4j/langchain4j-examples)](https://github.com/langchain4j/langchain4j-examples) (only for "big" features)
- [ ] I have added/updated [[Spring Boot starter(s)](https://github.com/langchain4j/langchain4j-spring)](https://github.com/langchain4j/langchain4j-spring) (if applicable)

- [X] I have added a `DynamoDbEmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or `EmbeddingStoreWithFilteringIT`
- [X] I have added a `DynamoDbEmbeddingStoreRemovalIT` that extends from `EmbeddingStoreWithRemovalIT`

### DIFF
--- a/embedding-stores/langchain4j-community-dynamodb/pom.xml
+++ b/embedding-stores/langchain4j-community-dynamodb/pom.xml
@@ -1,0 +1,189 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-community</artifactId>
+        <version>1.19.0-beta29-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-community-dynamodb</artifactId>
+    <name>LangChain4j :: Community :: Integration :: Amazon DynamoDB</name>
+
+    <properties>
+        <!--
+            Override parent's AWS SDK version.
+            DynamoDB native vector search (the SearchVectors operation and vector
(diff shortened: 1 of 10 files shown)


Martin7-1 (approved):
@riferrei Thank you! All changes look good to me

anandfresh (approved):

PR #773, opened 2026-08-22
Add Serply web search engine

### LINKED ISSUE #772: [FEATURE] Add Serply web search engine
## Is your feature request related to a problem? Please describe.

The community repo currently ships two `WebSearchEngine` implementations: SearXNG (self-hosted) and DuckDuckGo (keyless scraping). There is no hosted Google SERP option here, so agents that need Google-quality results with a simple API key have to pull in a search engine from outside the langchain4j ecosystem or write their own client.

## Describe the solution you'd like

A `langchain4j-community-web-search-engine-serply` module implementing `WebSearchEngine` on top of the Serply API (https://serply.io, Google SERP with a free tier). Same shape as the existing web-search-engine modules: a small client over `langchain4j-http-client`, a builder-based engine class, unit tests plus an integration test extending `WebSearchEngineIT` (skipped when `SERPLY_API_KEY` is unset), and registration in the root pom and BOM.

## Describe alternatives you've considered

Using the existing DuckDuckGo module (different result quality, no API contract) or the SearXNG module (requires operating an instance). A Serply engine could also live in a private codebase, but it fits the pattern this repo already establishes for pluggable web search engines.

## Additional context

I have a working implementation with passing unit and live integration tests, and will open a draft PR referencing this issue. Disclosure: I work with Serply.

googio:
## Issue
Closes #772

## Change

Adds a `langchain4j-community-web-search-engine-serply` module: a `WebSearchEngine` implementation backed by the [Serply](https://serply.io) Google SERP API, following the shape of the existing SearXNG and DuckDuckGo modules.

- `SerplyClient` calls `GET /v1/search` via the pluggable `langchain4j-http-client` (JDK client by default), with an explicit `User-Agent` header (Serply sits behind Cloudflare, which rejects requests without one) and a 2-retry policy.
- `SerplyWebSearchEngine` exposes the usual builder (`apiKey`, `baseUrl`, `timeout`, `logRequests`, `logResponses`), maps organic results into `WebSearchOrganicResult`, honors `maxResults` via the `num` query param, and surfaces Serply's "People Also Ask" data under the `relatedQuestions` key of `searchMetadata()`.
- Unit tests cover the response mapping (organic results, related-questions metadata, results missing title/link are skipped). `SerplyWebSearchEngineIT` extends the shared `WebSearchEngineIT` and is skipped when `SERPLY_API_KEY` is unset, so CI stays green without a key.
- Module registered in the root `pom.xml` and the BOM; revapi is configured with the same external-class ignores the DynamoDB module uses for core types exposed in the API.

## Testing

```
./mvnw verify -pl web-search-engines/langchain4j-community-web-search-engine-serply
```

with `SERPLY_API_KEY` set: unit tests 3/3 and integration tests 2/2 green against the live API. `spotless:check` passes.

Serply support is optional; nothing changes for users who do not add this module, and the integration tests are skipped when `SERPLY_API_KEY` is unset.

Disclosure: I work with Serply. Happy to adjust scope, naming, or drop this entirely if it isn't a direction you want for the project.

## General checklist


Martin7-1:
Thanks for the contribution (and for the affiliation disclosure)! Overall this is a clean, well-shaped module — but a few things need attention.

**What I already did on this branch:**
- Rebased onto latest `main` (no conflicts).
- Fixed the module `pom.xml` parent version: `1.20.0-beta30-SNAPSHOT` → `1.21.0-beta31-SNAPSHOT`. This one was non-trivial: with the mismatched parent, Maven silently resolved a **stale parent from the local repo** and everything "passed", but a clean CI checkout would fail outright.
- Removed the inline revapi `<analysisConfiguration>` block from the module pom (see item 1 below). After removal, `mvn verify` passes against the parent's default revapi config — confirming it was dead weight.
- All amended into the single commit; verified with `mvn test` / `mvn verify -DskipSerplyITs=true`.

## What checks out ✅

- Root `pom.xml` + BOM registration — actually present, verified.
- Sibling conventions: `<name>`, `<licenses>`, `${langchain4j.*.version}` properties, `skipSerplyITs` + failsafe config, package layout, `*Test`/`*IT` naming, `WebSearchEngineIT` contract base, `Assumptions`-based skip when `SERPLY_API_KEY` is unset.
- Client details as claimed: `GET /v1/search`, `X-Api-Key` header (matches Serply's documented auth), `User-Agent` for Cloudflare, `maxRetries(2)`, correct null-guard on `num`.
- No key leakage: `HttpRequestLogger` masks `api-key` headers, and the builder keeps no `toString` with the key.

## Needs fixing

1. **`totalResults` semantics.** `WebSearchInformationResult.from(getOrDefault(response.getTotal(), (long) results.size()))` — the core contract says *total pages found by the engine*, but the fallback silently reports the page size. Also I couldn't find public evidence that Serply returns a top-level `total` field; if it doesn't, every response misreports. The second unit-test fixture (no `total`) locks this wrong semantics in. `pageNumber` is never populated either, even though the API supports `num`/`start` pagination and `WebSearchRequest.startPage()` exists.
2. **Missing snippet → `""` vs. the shared IT.** `getOrDefault(result.getDescription(), "")` combined with `WebSearchEngineIT.should_search`'s `assertThat(snippet()).isNotBlank()` assertion makes the live IT flaky whenever Google returns a description-less result. Core tolerates a null snippet — suggest passing through null.
3. **Rank dropped when only `realPosition` is present.** The mapping only reads `position`, but Serply organic results frequently omit it (several public integrations rely on `realPosition`). Consider falling back.
4. **Retry/error handling wastes time on auth failures.** The retry policy wraps all throwables, so a 401/403 or Cloudflare challenge burns two ~1s delayed retries before surfacing a raw `HttpException` with no status/body context. Suggest retrying only on retryable statuses (429/5xx) and failing fast with a mapped message otherwise.

## Minor / judgement calls

- `hasValue(String)` re-implements `dev.langchain4j.internal.Utils.isNullOrBlank`, which the same file already uses (`getOrDefault` static import).
- `withApiKey(String)` static factory: unused in-module, no sibling offers it — it's a permanent public API surface; keep only if intentional.
- `OrganicResult.metadata` is deserialized but never read — dead code until mapped.
- Two top-level classes (`SerplyWebSearchResponse` + `OrganicResult`) share one file; searxng splits `SearXNGResult`/`SearXNGResponse` into separate files.
- `relatedQuestions` passthrough is beyond issue #772's scope, but it's transparently documented — fine to keep, just remember it needs docs when the documentation checklist item is picked up.

Could you take a look at items 1–4? Happy to re-review after.

googio:
Thanks for the rebase and the parent-version fix, and for catching the stale-parent resolution. Pushed 8c901f34 on top of your amended commit with items 1 to 4 plus the minor points.

1. totalResults / pageNumber. I checked the live API: every `/v1/search` response carries a top-level `total` key, but it is `null` for regular web searches, so there is nothing reliable to report. The engine now uses `total` when it is present and otherwise falls back to the result count the way the Tavily engine in core does, but it no longer does so silently: the information metadata carries `totalResultsEstimated=true` in that case. `pageNumber` is populated from `startPage`, and the client sends the matching zero-based `start` offset (`(startPage - 1) * pageSize`, page size being `maxResults` or Serply's default of 10). Verified against the API that `start` changes the page returned.
2. Missing description is passed through as a null snippet, no more `""`.
3. Rank falls back to `realPosition` when `position` is absent (it is camel-cased in the API, so it needs an explicit `@JsonProperty`).
4. Errors: `HttpException` is mapped to the core typed exceptions with the status and (truncated) body in the message. 401/403 become `AuthenticationException`, other 4xx `InvalidRequestException`, both non-retriable and fail on the first attempt. 429 and 5xx map to `RateLimitException` / `InternalServerException` and go through the retry policy. An unparseable body is also not retried.

Minor points: `withApiKey` and the unread `metadata` field are gone, `hasValue` is replaced by `Utils.isNotNullOrBlank`, and `OrganicResult` moved to its own file as `SerplyOrganicResult`. `relatedQuestions` stays and will get a line in the docs when that item is picked up.

Tests: `SerplyClientTest` (8 tests) drives the client through a stub `HttpClient` and covers the request shape, fail-fast on 401/400, retries on 503/429 and the parse path; `SerplyWebSearchEngineTest` grew to 5 for the estimate flag, the `realPosition` fallback and the null snippet. `mvn verify` (spotless, revapi) is green, and the two live ITs pass with a real key.

One note on item 2: the shared `WebSearchEngineIT.should_search` asserts a non-blank snippet, so a description-less result would still fail it with null just as it did with `""`. I have not seen one in the live runs so far. If you would rather have the engine drop such results the way searxng does, that is a one-line filter change; say the word.

googio:
Rebased onto current `main` (f2ce0424) to pick up your f3943623. Head 8c901f34 -> d2946f5c.

The red `java_build (21)` on the previous head was not this module. It was `AlloyDBEmbeddingStoreRemovalIT` failing twice with "ids cannot be null or empty". That was attempt 2 of run 34493713584, which started 2026-09-13 13:41:52 UTC and finished at 13:54:45 UTC. Your commit "fix: make removeAll(ids) a silent no-op for null/empty input in remaining embedding stores" landed on `main` at 14:30:53 UTC, about 36 minutes after that job finished, so the build could not have contained it. `java_build (17)` and `java_build (25)` were green on the same attempt.

The rebase carries no content change for this PR. The diff against `main` is identical to the previous head apart from blob hashes, and the module parent version still matches the root pom at 1.21.0-beta31-SNAPSHOT.

One thing worth flagging: the rebase rewrote the commit you amended (2d8d90aa -> 89f382b9). Your parent-version fix is preserved in it, but the committer stamp is now mine as a side effect of the rebase. If you would rather I merge `main` in than rebase, say so and I will restore it.

PR #774, opened 2026-08-28
Add langchain4j-community-web-search-engine-brave module

takanori-ugai:
## Issue
  
 Closes #

## Change
  
  - Added the `langchain4j-community-web-search-engine-brave` module under `web-search-engines/`.
  - Implemented `BraveWebSearchEngine` using the Brave Web Search API with support for:
  - Custom API keys and custom base URLs.
  - Request parameter configurations (e.g., search language, country, freshness, safe search, spellcheck).
  - Secret masking for sensitive parameters in logs.
  - Placed all sources and tests under the package `dev.langchain4j.community.web.search.brave`.
  - Added comprehensive unit tests and integration tests (`BraveWebSearchEngineTest`, `BraveWebSearchEngineIT`,  `SecretMaskingTest`).

## General checklist
  
  - [ ] I have added an example in the [examples repo](https://github.com/langchain4j/langchain4j-examples) (only  for "big" features)

  - [ ] I have added a `{NameOfIntegration}EmbeddingStoreIT` that extends from either `EmbeddingStoreIT` or   `EmbeddingStoreWithFilteringIT`

  - [ ] I have manually verified that the `{NameOfIntegration}EmbeddingStore` works correctly with the data  persisted using the latest released version of LangChain4j

### DIFF
--- a/web-search-engines/langchain4j-community-web-search-engine-brave/pom.xml
+++ b/web-search-engines/langchain4j-community-web-search-engine-brave/pom.xml
@@ -1,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <parent>
+        <groupId>dev.langchain4j</groupId>
+        <artifactId>langchain4j-community</artifactId>
+        <version>1.21.0-beta31-SNAPSHOT</version>
+        <relativePath>../../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>langchain4j-community-web-search-engine-brave</artifactId>
+    <packaging>jar</packaging>
+
+    <name>LangChain4j :: Web Search Engine :: Brave</name>
+    <description>Implementation of Brave Search API for LangChain4j</description>
+
+    <dependencies>
+
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-core</artifactId>
+            <version>${langchain4j.core.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-http-client</artifactId>
+            <version>${langchain4j.http-client-jdk.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-http-client-jdk</artifactId>
+            <version>${langchain4j.http-client-jdk.version}</version>
+            <scope>runtime</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
+
+        <!-- test dependencies -->
+
+        <!-- Visibility for WebSearchEngineIT -->
+        <dependency>
+            <groupId>dev.langchain4j</groupId>
+            <artifactId>langchain4j-core</artifactId>
+            <version>${langchain4j.core.version}</version>
+            <classifier>tests</classifier>
+            <type>test-jar</type>
+            <scope>test</scope>
+        </dependency>
+
+    </dependencies>
+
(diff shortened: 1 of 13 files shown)


Martin7-1:
Thanks for the contribution! 👋 A few housekeeping notes first, then the review.

**What I already did on this branch:**
- Rebased onto latest `main` (no conflicts).
- Amended the commit to fix the module `pom.xml` version wiring: parent `1.20.0-beta30-SNAPSHOT` → `1.21.0-beta31-SNAPSHOT`, and hardcoded `1.20.0-SNAPSHOT` deps → `${langchain4j.core.version}` / `${langchain4j.http-client-jdk.version}` to match sibling modules. Verified with `mvn test-compile`.

## Blocking

1. **Module is not registered.** The diff only contains files under the new module — it is missing from `<modules>` in the root `pom.xml` and from `langchain4j-community-bom/pom.xml`. As-is, the module is never built or tested in CI. The PR checklist claims this was done, but it isn't in the diff. Please add both.
2. **`safesearch` semantics.** `WebSearchRequest` from core defaults `safeSearch=true`, so *every* request (even a plain `search("...")`) sends `safesearch=strict`, silently overriding Brave's documented `moderate` default — and `moderate` is unreachable. Suggest mapping `Optional`-style explicitly and only emitting the param when the caller set it.
3. **Offset / page semantics.** `offset` is computed as a page index, but Brave defines it as a 0-based starting position. With `count < 20` (e.g. `maxResults=5, startPage=3`) the fetched slice is wrong. Also when the offset is clamped, the response still reports the requested page (e.g. `startPage=25` → offset 9, reported page 25), and `totalResults` is filled with the page size (`results.size()`) instead of the total found.
4. **Error handling.** No non-2xx / rate-limit handling; `BraveJsonUtils` wraps parse failures in a bare `RuntimeException` with no status code or response body, which makes API failures very hard to diagnose.

## Spec gaps

- The PR description claims support for *freshness* and *spellcheck*, but neither is a first-class parameter — only `additionalParams` pass-through exists, and `spellcheck` is not even mentioned in the class javadoc.
- `extraSnippets` is parsed but never mapped to the response, and `extra_snippets=true` is never requested — dead field.
- **Secret masking is narrower than claimed.** Only the builders' `toString()` masks the key. Log masking of the `X-Subscription-Token` header relies entirely on core `HttpRequestLogger`'s keyword-based redaction, and sensitive values embedded in a custom `baseUrl` or in `additionalParams` end up in request logs verbatim. `SecretMaskingTest` currently only asserts on `toString()`.

## Scope creep / cleanups (non-blocking)

- Unused public constructor `BraveWebSearchEngine(String, String, Duration)` duplicating the builder.
- "Legacy response with top-level `results`" dual-shape support — Brave's API only defines `web.results`.
- `BraveWebSearchResponse.getType()` and `BraveQuery.getOriginal()/getAltered()` are parsed but never read.
- `revapi.json` with 4 blanket ignore rules — sibling search-engine modules have none.
- Duplicated code: `BraveClientBuilder` vs `BraveWebSearchEngineBuilder` carry six identical fields/methods plus the same `toString` masking logic — worth extracting.
- Magic strings: `"strict"`/`"off"` if-cascade for `safesearch`, metadata keys `"page_age"` / `"moreResultsAvailable"` — consider a small enum / constants.
- `private BraveJsonUtils() throws InstantiationException` is an unusual util-ctor pattern.
- Minor convention alignment with siblings: `<licenses>` block, `<name>` style (`LangChain4j :: Community :: Integration :: Brave`).

## Positives

Package layout, `*Test`/`*IT` naming, extending `WebSearchEngineIT`, and the unit-test coverage (mapping, clamping, response shapes) all follow repo conventions and look meaningful — they just don't run anywhere until item 1 is fixed.

Could you address the blocking items (registration first)? Happy to re-review after.

takanori-ugai:
Thanks for the thorough review! I’ve addressed the requested items in commit 38c5188.

  - Registered the module in the root pom.xml and community BOM.
  - Fixed SafeSearch so plain requests preserve Brave’s default moderate behavior; false maps to off.
  - Corrected offset/page calculations, clamping, and effective-page metadata.
  - Added status-code, rate-limit, response-body, and JSON parse error details.
  - Added explicit request handling for freshness, spellcheck, and extra snippets, including response mapping.
  - Expanded secret masking to URLs, query parameters, headers, and logs.
  - Removed legacy response parsing, unused fields/constructors, blanket Revapi ignores, and duplicated builder code.
  - Added/updated tests and aligned formatting.

 Verification completed successfully:

 13 tests passed
 Spotless check passed

 Happy to have this re-reviewed.

PR #792, opened 2026-09-14
fix: terminate streaming router subscriptions on invalid demand

### LINKED ISSUE #791: [BUG] StreamingModelRouter throws for non-positive demand instead of signaling onError
## Describe the bug

At community main commit `f2ce042433cbd8f946665a41d6b9dc3fa1039b8c`, `StreamingModelRouter.FailoverSubscription.request(long)` throws `IllegalArgumentException` for `n <= 0` instead of notifying the subscriber through `onError`.

I reproduced this through the public `router.chat(ChatRequest).subscribe(...)` entry point with deterministic local tests for both `0` and `-1`:

- Inside `onSubscribe`, the exception escapes `subscribe()`, no `onError` is delivered, and the delegate has not been invoked.
- After `subscribe()` returns with an active, non-terminating delegate, the exception escapes `request()`, and no `onError` is delivered.

A separate controlled-upstream probe with `request(0)` observed no upstream cancellation. A subsequent positive request could still receive an event. If the upstream failed before any event, the router still attempted another route. The invalid request itself did not directly trigger that failover.

## Log and Stack trace

The original 8 `StreamingModelRouterTest` tests passed. Four new parameterized cases failed against unchanged production code. Each reported an unexpected directly thrown exception and an error-signal count of zero instead of one.

Representative stack frames for the onSubscribe case:

```text
java.lang.IllegalArgumentException: non-positive request: 0
  at StreamingModelRouter$FailoverSubscription.request(StreamingModelRouter.java:249)
  at StreamingModelRouterTest$DemandCollector.onSubscribe(...)
  at StreamingChatModel$2$1.onSubscribe(StreamingChatModel.java:216)
  at StreamingModelRouter$FailoverSubscription.start(StreamingModelRouter.java:186)
  at StreamingChatModel$2.subscribe(StreamingChatModel.java:210)
```

## Minimal Example to Reproduce

Add the following test method to the existing `StreamingModelRouterTest` class in `models/langchain4j-community-model-router/src/test/java/dev/langchain4j/model/router/StreamingModelRouterTest.java`:

```java
@Test
void shouldSignalErrorOnNonPositiveDemand() {
    ScriptedStreamingModel model = new ScriptedStreamingModel(subscriber -> {}, null);
    StreamingModelRouter router = StreamingModelRouter.builder()
            .addRoutes(model)
            .routingStrategy(fixedIndexStrategy(0))
            .build();
    List<Throwable> errors = new java.util.ArrayList<>();
    Subscriber<ChatModelStreamingEvent> subscriber = new Subscriber<>() {
        @Override
        public void onSubscribe(Subscription subscription) {
            subscription.request(0);
        }
        @Override
        public void onNext(ChatModelStreamingEvent event) {}
        @Override
        public void onError(Throwable error) { errors.add(error); }
        @Override
        public void onComplete() {}
    };
    org.junit.jupiter.api.Assertions.assertAll(
            () -> org.junit.jupiter.api.Assertions.assertDoesNotThrow(
                    () -> router.chat(REQUEST).subscribe(subscriber)),
            () -> assertEquals(1, errors.size()));
}
```

This method reuses the existing fixture methods, imports and `REQUEST` from that test class. From the repository root, run:

```sh
mvn -f models/langchain4j-community-model-router/pom.xml \
  -Dtest=StreamingModelRouterTest#shouldSignalErrorOnNonPositiveDemand test
```

The two assertions fail: an exception escapes `subscribe()`, and the error-signal count is 0. No model credentials are needed. Changing `0` to `-1` exercises the other reproduced value. For the later-request case, store the subscription in `onSubscribe` without requesting, then call `request(0)` after `subscribe()` returns.

## Expected behavior

An active subscription receiving non-positive demand should notify its subscriber with `onError(IllegalArgumentException)`, and `request` should return normally when the subscriber callbacks return normally. The invalid-demand termination should not start or fail over to another model; an existing upstream should be cancelled and subsequent signals handled consistently with termination.

References: [Java 17 Flow.Subscription](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/concurrent/Flow.Subscription.html), [Reactive Streams rules 3.9 and 3.16](https://github.com/reactive-streams/reactive-streams-jvm/blob/master/README.md).

## Please complete the following information

- Community version: main at `f2ce042433cbd8f946665a41d6b9dc3fa1039b8c`, `1.21.0-beta31-SNAPSHOT`
- Component: `langchain4j-community-model-router`
- Core dependency: `1.21.0-SNAPSHOT`, resolved timestamp `20260911.135020-34`
- Java: Microsoft OpenJDK 21.0.10, compilation targeting Java 17
- Maven: 3.9.12
- Spring Boot: not used

## Additional context

Hi @Martin7-1, I saw that your [self-review on #776](https://github.com/langchain4j/langchain4j-community/pull/776#pullrequestreview-5113157773) already noted this behavior as non-blocking. I am opening this issue to track the reproduced contract mismatch and would like to contribute the fix.

If this follow-up is welcome and nobody is already working on it, could you assign this issue to me, or confirm that I can proceed? I would prepare a focused Draft PR with regression tests, keeping Java 17 compatibility and the existing normal routing/failover behavior.

My proposed scope is invalid-demand termination in `StreamingModelRouter`, including the necessary upstream cancellation and terminal-signal handling. The broader native async work in #786 would remain a separate task. No production fix has been implemented yet; I would appreciate confirmation of this scope before starting it.

Shay-Deng:
## Issue

Fixes #791

## Change

`StreamingModelRouter` threw from `request(n <= 0)` instead of signaling `onError`. From `onSubscribe`, that exception escaped `subscribe()`; after subscription setup, the upstream remained active.

Invalid demand now selects a single terminal `IllegalArgumentException`, cancels an existing or late-registering upstream, and stops further routing and signal forwarding. Requests after cancellation or termination are ignored. If `onNext` is in progress, the terminal signal is deferred until that callback returns, including reentrant and cross-thread requests. Downstream callbacks and upstream cancellation run outside the subscription monitor.

The change is confined to the router's private subscription implementation and regression tests. Public APIs, dependencies and routing strategies are unchanged.

Validation:

- Invalid-demand regression tests fail against the original implementation and pass with the fix.
- `StreamingModelRouterTest`: 20 tests pass, covering finite demand across failover, cancellation callbacks, late registration/signals, terminal idempotence, and deterministic callback ordering.
- Entire model-router module: `mvn -f models/langchain4j-community-model-router/pom.xml -Pspotless clean verify` passes on JDK 17 and JDK 21; 33 tests, 0 failures/errors, 2 existing Azure credential-gated tests skipped.
- Spotless, Javadoc and Revapi checks pass with the existing project configuration.

## General checklist

Notes: The new tests exercise the public reactive entry point with deterministic test models. No new service integration tests were added, so the combined unit-and-integration checkbox is left unchecked. Only the changed module was run; the existing credential-gated Azure tests were skipped. Documentation and examples can follow after review if requested; Spring Boot changes are not applicable.

Not applicable.

Not applicable.

Not applicable.


Martin7-1 (approved):
@Shay-Deng Thank you!

PR #799, opened 2026-09-21
fix: fixing Xinference ITs

diegoberriosr:
## Issue

Closes none, but related to #745 

## Change

<div>
image
</div>
<br/>
Fixed some ITs in `langchain4j-community-xinference` that were failing due to various reasons described below

#### `XinferenceToolChatModelIT`
- Added `XINFERENCE_AUTH_ADVANCED=false` during the container setup (`XinferenceContainer.java`), otherwise most tests will fail if they follow that path in the newest image versions.
- `beforeEach()` has a null check for building the `chatModel` instance, with the current setup it defaults to `null` on the first test because `models()` runs while it still hasn't been instantiated.
- `should_fail_when_tool_returns_image_and_provider_does_not_support_it`:  This test was added during [this context](https://github.com/langchain4j/langchain4j/pull/4851) in the core repo. Xinference does not support tools that return images, so an exception is thrown when the tool's response contains content other than text (following the pattern of other integrations.
- `should_handle_tool_without_parameter`: `XinferenceInternalHelper.toParameters()` did not contemplate this case, so the test failed because an incorrect schema was being sent to the API.
- `should_execute_a_tool_then_answer`: I noticed this test was flaky, sometimes the model would output `{"location": "Paris", "format": "celsius"}`  breaking the assertion on the string, so I added de-serialization to avoid this issue. I also changed the assertion on the response's `toolExecutionRequest` to checking for emptiness instead of `null`, since the internal mapper returns empty lists in those cases.
- `should_not_execute_a_tool_and_tell_a_joke`: Just like the test above, the assertion was changed to `isEmpty()` to match the intent of `XinferenceInternalHelper`.

The rest of the tests in this class were basically failing because it seems like `qwen3` at `0_6B` could not handle nested schemas reliably, so I bumped the model in the test to `qwen3.5` and it fixed all of them but `should_execute_tool_with_list_of_POJOs_parameter` and `should_execute_tool_parameters` (they ended up being flaky or failed directly).

(I also modified some prompts and the test's tool schema a little bit to reduce the flakiness of some tests).

#### `XinferenceLanguageModelIT` & `XinferenceStreamingLanguageModelIT`

These ones were failing due to two reasons:

1. It would seem like the `Transformers` + `pytorch` config causes a bug in the container when the `/completions` endpoint is used. This doesn't happen with the `llama.cpp` config: 

```dev.langchain4j.community.model.xinference.client.XinferenceHttpException: {"detail":"[address=0.0.0.0:38605, pid=678] 'str' object has no attribute 'get'"}```

2. `should_generate_answer_and_return_token_usage_and_finish_reason_stop` and `should_stream_answer` were still flaky sometimes. I changed the prompt a little bit to try to force `北京` early in the sequence while still generating a long enough answer in order to be truncated.

<did>
Screenshot from 2026-09-20 20-45-23
</div>

## General checklist


pamod-madubashana (commented):
Bugfix review:

Summary: Fixes Xinference tool mapping - always `type:object`, `emptyMap()` for no-arg tools, fail-fast on non-text tool results. IT infra moves to qwen3.5 + auth fix.

1. New `UnsupportedFeatureException` for non-text tool results is breaking vs prior silent truncation - document in changelog.
2. `emptyMap()` vs `null` - confirm server accepts `{type:object, properties:{}}` for no-arg tools.
3. Prompt hardening (`CALL_ONE_TOOL...`) + `@Disabled` tests mask 2B-model flakiness. Add fast unit tests for `toXinferenceMessages`/`toParameters` so fix doesn't depend only on 10-min docker ITs.

Otherwise LGTM.

